PR #345, opened 2023-03-18
feat: add tailwind-converter

astahmer:
## what is this

currently a tailwind to CSS-in-JS converter

##  in the future it could : 
- keep the theme token values

atm: `bg-slate-100`  -> background: "#12345"
future: `bg-slate-100` -> background: "slate.100" (given slate.100 is defined in the `theme.colors`)

- also convert the tailwind config to the panda config equivalent
- atm it extracts (using box-extractor) class/className jsx attributes but in the future there could be a Combobox or some kind of inputs to specify which fn args to extract (think `cx`, `clsx`, `cn`, etc etc)

Screenshot 2023-03-18 at 23 34 15

## how does it work

using tailwind internals (1) we can directly use their config + function to convert to postcss AST and then walk this AST to convert to a CSS-in-JS object

```ts
// (1)
import { resolveMatches as resolveMatchesRaw } from 'tailwindcss/lib/lib/generateRules'
import { createContext as createContextRaw } from 'tailwindcss/lib/lib/setupContextUtils'
```

rough implementation is in `tailwind-converter/src/converter/tw.ts`

<hr />

as a side note, the current tailwind-converter directory was ported from a local PoC vite repo so some eslint/ts lint rules may throw, some logs might still be commented out, in summary this is not clean yet

### DIFF
--- a/packages/cli/app/src/pages/index.astro
+++ b/packages/cli/app/src/pages/index.astro
@@ -5,5 +5,5 @@ import { customDocs } from '../utils/custom-docs'
 ---
 
 <Layout title={customDocs.pageTitle}>
-  <Overview client:load />
+  <Overview client:only />
 </Layout>
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -577,6 +577,81 @@ importers:
       typescript: 4.9.5
       vite: 4.1.4
 
+  tailwind-converter:
+    specifiers:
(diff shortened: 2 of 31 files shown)


DmacMcgreg:
@segunadebayo @astahmer 
Did this get closed because there's going to be a tailwind preset instead? 
Seems like a killer feature to be able to just copy and paste tailwind examples from the net into a panda project.

PR #680, opened 2023-06-12
feat(config): incremental/chunks boolean

astahmer:
brings some perf improvements by not writing chunks file and re-using parser results with a panda.buildinfo.json file on initial startup

image


segunadebayo:
As we discussed, let's keep this on the back burner until we really need it.

PR #706, opened 2023-06-17
fix(readme): fix link to UnoCSS

Kocal:
Otherwise the link points to https://github.com/chakra-ui/panda/blob/main/unocss.dev


Kocal:
Duplicate of #704, sry

PR #707, opened 2023-06-17
fix: types and eslint warning in local projects

manucorporat:


anubra266:
Hi @manucorporat could you explain more what issue this solves.

segunadebayo:
@anubra266, the goal of the change is to allow adding the styled-system dir inside the `src` directory of a project.

The eslint-disable comment will allow eslint skip checking those files.

anubra266:
@segunadebayo we might want to do this for prettier then

PR #708, opened 2023-06-17
Fixed Uno CSS Redirection Link in Readme

Dhananjay-JSR:
Before : - It Used to Redirect to "https://github.com/chakra-ui/panda/blob/main/unocss.dev"
After Modification :- It Redirects to correct "https://unocss.dev/"


segunadebayo:
This is fixed already

PR #715, opened 2023-06-17
docs: add getting started with Vue

wobsoriano:
- Add Vue in the Getting Started section


website/pages/docs/getting-started/vue.mdx
```diff
@@ -0,0 +1,172 @@
...
+    ```vue filename="src/App.vue"
+    <script setup lang="ts">
+    import { css } from '../styled-system/css';
+    </script>
+
+    <template>
+      <div :class="css({ fontSize: '2xl', fontWeight: 'bold' })">Hello 🐼!</div>
+    </template>
+    ```
+  </Tab>
```
segunadebayo, line 155: Panda can only work in Vue JSX syntax since it parses only JS(X)/TS(X) files for now.

Let's remove this section and add a note about SFC in Vue, which is not supported for now.

wobsoriano: It works on SFC! Unless Im missing something?

I’ll try to build the app and check if I get an error. Either way, will update the PR later.

wobsoriano: Panda works 100% with SFC @segunadebayo. Do you know a specific Panda feature that is not supported?

Screenshot 2023-06-18 at 9 26 27 PM

anubra266: @wobsoriano could you check the browser and see if the styles were applied. They won't be.

paul-nallet: I've tried and it doesn't seem to work in my browser in build or dev. As @wobsoriano i saw the cli updates

segunadebayo: Yes, this is happening because our static analysis engine doesn't support Vue language syntax at the moment. We're going to support that in the future.

For now, I'd recommend putting the styles in a dedicated `<component>.css.ts` file and declaring the styles there, pretty much like CSS Modules.

This breaks co-location but we'll definitely improve the DX

wobsoriano: So, yeah. The styles weren't applied. How did I miss that? 😆 

And for the usage, I added a note about JSX being supported for now.

PR #795, opened 2023-06-22
fix(parser): factory with prop access should be like `css`

astahmer:
my bad I misunderstood how the the `styled.div({ ... })` usage, it should be like `css` not like `cva`

fixes this one https://github.com/chakra-ui/panda/pull/787

### DIFF
--- a/packages/parser/__tests__/output.test.ts
+++ b/packages/parser/__tests__/output.test.ts
@@ -481,10 +481,8 @@ describe('extract to css output pipeline', () => {
       })
 
       const Input = panda.input({
-        base: {
-          color: "blue.100",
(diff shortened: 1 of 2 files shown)


anubra266:
@astahmer there's no changeset

astahmer:
> @astahmer there's no changeset

i'll update this one https://github.com/chakra-ui/panda/blob/30f41e01153c01cd69c27e10a27a5a226cf9eb56/.changeset/famous-students-reflect.md from the previous PR when we have a final decision regarding that API

astahmer:
closing this one as i'll treat it in this one https://github.com/chakra-ui/panda/pull/786 with the others new use cases

PR #811, opened 2023-06-24
docs: persist package manager

isBatak:
## 📝 Description

- docs tabs state is shared between all the same tabs. 
- used neutral colors for tab styles

https://github.com/chakra-ui/panda/assets/4128883/3c5a6ef4-87d7-464b-8897-9b15cfdeb774


segunadebayo:
Thanks for these improvements @isBatak.

I'd prefer if we keep the previous styles for the tabs. I am trying to minimize the yellow even though I understand the intent is to keep it on brand.

isBatak:
> Thanks for these improvements @isBatak.
> 
> I'd prefer if we keep the previous styles for the tabs. I am trying to minimize the yellow even though I understand the intent is to keep it on brand.

Then I would remove all the yellow from the docs and keep this blue color. It really bothers me when the half of stuff uses yellow as the `accent` color and the other half uses blue. It's inconsistent.

segunadebayo:
I didn't mean, "remove all the yellow from the docs."

For me, it's more about visual balance than consistency. I did like the outline version of the tabs we had before.

We can explore neutral colors with the previous version if the color bothers you.

PR #812, opened 2023-06-24
playground: improvements

anubra266:
- Fix token method
- Allow editing config


segunadebayo:
Here's a screenshot of the error from Vercel

CleanShot 2023-06-25 at 20 23 42@2x

Issue #817, opened 2023-06-25
Breakpoints starting with a number aren't working (2xl, etc)


IHIutch:
### Description

When I use 2xl breakpoint, I expect my styles to render, but instead my 2xl breakpoint styles are never rendered

https://stackblitz.com/edit/stackblitz-starters-3siapy?file=pages%2Findex.js

### Link to Reproduction

https://stackblitz.com/edit/stackblitz-starters-3siapy?file=pages%2Findex.js

### Steps to reproduce

1. Add styles to a break point starting with a number (`2xl`, `3xl`, etc)
2. Can be within a style attribute or breakpoint attribute
```
className={css({
  bg: {
    '2xl': 'orange.500',
  },
})}
// or
className={css({
  '2xl: {
    bg: 'orange.500',
  },
})}
```
3. Styles never render at specified breakpoint, although you can see them generated in your CSS
```
@media screen and (min-width: 1536px) {
    .\2xl\:bg_orange\.500 {
      background: var(--colors-orange-500)
    }
}
```

### JS Framework

React 

### Panda CSS Version

0.4.0

### Browser

Brave Version 1.52.126 Chromium: 114.0.5735.133 (Official Build) (arm64)

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

For what its worth, I checked out how [Tailwind handles this](https://play.tailwindcss.com/6GA4QVipo5) and it prefixes a `3` to the style name. Not sure why, though.

```
@media (min-width: 1536px) {
  .\32xl\:bg-orange-400 {
    --tw-bg-opacity: 1;
    background-color: rgb(251 146 60 / var(--tw-bg-opacity));
  }
}
```

IHIutch:
After some cursory reading, it looks like `.\32` is used because 32 is the unicode position of `2`. Today I learned.

However, I don't know what tooling is used to generate their (or Panda's) classNames, otherwise, I'd happily open a PR!

- https://www.w3.org/International/questions/qa-escapes#css_identifiers
- https://mathiasbynens.be/notes/css-escapes#css

segunadebayo:
Thanks for such a detailed breakdown. TIL

I've gone ahead to update the escape function. I'll release an update shortly.

IHIutch:
Ah, ha. I was actually digging into this myself. But I'll post what Ive been working on in case its useful

PR #819, opened 2023-06-25
Fix classnames starting with number

IHIutch:
Fixes https://github.com/chakra-ui/panda/issues/817

There are two ways to go here: (at least, from what I came up with)

Option A just checks to see if the className starts with a number and then converts that to safe unicode. 

Option B uses [`cssesc` library](https://github.com/mathiasbynens/cssesc) which is what [postcss-selector-parser](https://github.com/postcss/postcss-selector-parser/blob/09d5d48189d0c4e5d02c71191608153abd8e0c61/src/selectors/className.js#L15) in their classNames function.

This might let you replace the `esc` function altogether, but Im not exactly sure what else the regex handles.

---

I also added tests for 2xl breakpoint, but I think they were already passing so 🤷


segunadebayo:
Thanks for digging into this @IHIutch.

As you may have noticed, I also pushed a fix for this yesterday.

Here's the code:
https://github.com/chakra-ui/panda/blob/main/packages/shared/src/esc.ts

Here's the test:
https://github.com/chakra-ui/panda/blob/e16c06091ae6666b48162fef9dc740aa818361a0/packages/shared/__tests__/esc.test.ts#L8

If it becomes problematic, we'll use Option B, the postcss-selector-parser instead.

IHIutch:
Works for me, thanks

PR #823, opened 2023-06-26
docs: preset code examples

qin-guan:
No GitHub issue

## 📝 Description

In current documentation, theme object passed to `definePreset` is not wrapped by `theme` key. It should be wrapped as shown in example below from `preset-panda`

https://github.com/chakra-ui/panda/blob/main/packages/preset-panda/src/index.ts#L10

N/A

## 🚀 New behavior

N/A

No


qin-guan:
Thanks @segunadebayo! 🎉 Really appreciate the amazing work you’re doing for the Chakra ecosystem 🙏🏻!

Issue #813, opened 2023-06-24
Svelte extraction doesn't include context module script


pawelblaszczyk5:
### Description

In Svelte there is a concept of additional block of script, that is shared between component instances (https://svelte.dev/docs/svelte-components#script-context-module). It seems like extraction doesn't include the styles declared through this script. I copied yours Svelte sandbox and added my simple example.

### Link to Reproduction

https://github.com/pawelblaszczyk5/svelte-panda-repro

### Steps to reproduce

1. Clone the repository
2. Run it
3. Go to the default +page.svelte file
4. See that last styles aren't applied

### JS Framework

Svelte (TS)

### Panda CSS Version

0.4.0

### Browser

n/a

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

segunadebayo:
I merged a PR from @astahmer that fixed this for Vue and Svelte.

pawelblaszczyk5:
> I merged a PR from @astahmer that fixed this for Vue and Svelte.
> 
> We’ll release an update pretty soon

Isn't his PR already in the latest version? Or is there a way to check it with a pre release or something?

astahmer:
that [PR](https://github.com/chakra-ui/panda/commit/8991b1e4f064df42abb29750267a3bcd80a3f4a5) included svelte but it seems the default svelte parser doesn't care about Typescript, so i'll have to tweak [the current implementation](https://github.com/chakra-ui/panda/blob/3a87cff827b3a5f310768b053ddd66d3b24a7605/packages/parser/src/svelte-to-tsx.ts#L6) to make use of [svelte-preprocessor](https://github.com/sveltejs/svelte-preprocess/blob/main/docs/preprocessing.md#typescript) for that 

tho, without Typescript you should be fine
you can try it with [the latests versions published](https://www.npmjs.com/package/@pandacss/dev?activeTab=versions) (automatically when pushed on main, you could use `latest` as well I suppose)

pawelblaszczyk5:
So maybe we could reopen this issue until it's solved for TypeScript as well?

pawelblaszczyk5:
@astahmer I'd be open to contribute the svelte-preprocess part if it's something you're interested in. Just let me know and I could work on it 😇

astahmer:
feel free to do so ! basically the only requirement is that the process remains synchronous and that it doesn't break anything

I'm available on discord if you need help

pawelblaszczyk5:
[...]

PR #827, opened 2023-06-26
fix(parser): svelte with TS / script context / multiple script

astahmer:
Closes https://github.com/chakra-ui/panda/issues/813

## 📝 Description

Fix svelte parsing 

- `<script context="module">` content would not be extracted properly
- Typescript usage in a `<script>` would make the `svelte.parse(code)` crash
- using multiple <script> wasn't working properly either

## 🚀 New behavior

fixed

no

## 📝 Additional Information

implementation-wise, I removed the `svelte.parse` cause we can't use `svelte-preprocess` either because of its async-ness

surprisingly it seems ts-morph is parsing just fine in svelte-specific syntax such as `{#if xxx}` `{:catch}` etc, so we just remove the `<script xxx>` & `<style>` tags with regexes

### DIFF
--- a/.changeset/popular-foxes-deny.md
+++ b/.changeset/popular-foxes-deny.md
@@ -1,0 +1,5 @@
+---
+'@pandacss/parser': patch
+---
+
+Fix svelte parsing when using Typescript or <script context="module"> or multiple <script>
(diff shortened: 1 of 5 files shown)


pawelblaszczyk5:
You've took a really interesting approach, awesome idea and job, thank you for this 🎉

segunadebayo:
This is such an elegant solution @astahmer. Thanks for working on this 💛

PR #853, opened 2023-06-27
fix: replace where with is condition to apply specificity

blechatellier:
image

When nesting an `_invalid` condition, css are being overwritten because the `:where` selector is not specific.

Replacing all `:where` with `:is` condition to apply specificity to nested properties.

[Discussion on discord](https://discord.com/channels/1118988919804010566/1120305029056831548/1122950787090366617)


omarkhatibco:
Hey @blechatellier, 
Thanks for this PR, could you please update the docs as well?
https://panda-css.com/docs/concepts/conditional-styles#reference

blechatellier:
@omarkhatibco thanks for pointing out, it has been updated.

PR #922, opened 2023-07-03
fix(types): allow `!` with ColorPalette value

astahmer:
## 📝 Description

Fix ColorPalette typings by allowing `!` with the value to make it important

Screenshot 2023-07-03 at 10 21 03

type would throw, runtime would be fine

no


omarkhatibco:
fyi, It's working on textStyle but I got a type error as well

PR #933, opened 2023-07-04
feat(generator): export isCssProperty from /jsx

astahmer:
## 📝 Description

Export `isCssProperty` helper function from styled-system/jsx
It is mostly useful to make a wrapper around `styled` factory function, so split properties between style props & custom ones

no


packages/generator/src/artifacts/js/is-valid-prop.ts
```diff
@@ -10,5 +11,11 @@ export function generateisValidProp(ctx: Context) {
   )
   return {
     js: content,
+    ts: outdent`
```
segunadebayo, line 14: To be consistent with others

```suggestion
    dts: outdent`
```

Issue #930, opened 2023-07-04
feature request: Extend the Transform utilities to accept negative values.


omarkhatibco:
### Description

Hey, 

this feature is about this file [here](https://github.com/chakra-ui/panda/blob/main/packages/preset-base/src/utilities/transforms.ts#L37C3-L37C13)

### Problem Statement/Justification

the `translateX` and `translateY` fns should accept negative values as well.
in reality negative values also used as much as positive one

### Proposed Solution or API

in this [line](https://github.com/chakra-ui/panda/blob/946a38d966887e852d02050befbaa9188b1d0b19/packages/preset-base/src/utilities/transforms.ts#L37)

```diff

  translateX: {
    shorthand: 'x',
    className: 'translate-x',
    values(theme) {
      return {
        ...theme('spacing'),
        '1/2': '50%',
        '1/3': '33.333333%',
        '2/3': '66.666667%',
        '1/4': '25%',
        '2/4': '50%',
        '3/4': '75%',
        full: '100%',
+       '-1/2': '-50%',
+       '-1/3': '-33.333333%',
+       '-2/3': '-66.666667%',
+       '-1/4': '-25%',
+       '-2/4': '-50%',
+       '-3/4': '-75%',
+       '-full': '-100%',
      }
    },
    transform(value) {
      return {
        '--translate-x': value,
      }
    },
  },
  translateY: {
    shorthand: 'y',
    className: 'translate-y',
    values(theme) {
      return {
        ...theme('spacing'),
        '1/2': '50%',
        '1/3': '33.333333%',
        '2/3': '66.666667%',
        '1/4': '25%',
        '2/4': '50%',
        '3/4': '75%',
        full: '100%',
+       '-1/2': '-50%',
+       '-1/3': '-33.333333%',
+       '-2/3': '-66.666667%',
+       '-1/4': '-25%',
+       '-2/4': '-50%',
+       '-3/4': '-75%',
+       '-full': '-100%',
      }
    },
    transform(value) {
      return {
        '--translate-y': value,
      }
    },
  },

```

PR #934, opened 2023-07-04
feat(preset-base): translateXY negative fraction values

astahmer:
Closes https://github.com/chakra-ui/panda/issues/930

## 📝 Description

Add negative fraction values to `translateX` and `translateY` utilities

no

### DIFF
--- a/packages/preset-base/src/utilities/transforms.ts
+++ b/packages/preset-base/src/utilities/transforms.ts
@@ -1,5 +1,18 @@
 import type { UtilityConfig } from '@pandacss/types'
 
+const positiveFractions = {
+  '1/2': '50%',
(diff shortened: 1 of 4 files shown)


packages/preset-base/src/utilities/transforms.ts
```diff
@@ -1,5 +1,18 @@
 import type { UtilityConfig } from '@pandacss/types'
 
+const positiveFractions = {
+  '1/2': '50%',
+  '1/3': '33.333333%',
+  '2/3': '66.666667%',
+  '1/4': '25%',
+  '2/4': '50%',
+  '3/4': '75%',
```
segunadebayo, line 9: Let's add "full" to this list as well

PR #944, opened 2023-07-05
types: definePattern module override missing extends

astahmer:
## 📝 Description

Fix `definePattern` module overriden type, was missing an `extends` constraint which lead to a type error:

```
styled-system/types/global.d.ts:14:58 - error TS2344: Type 'T' does not satisfy the constraint 'PatternProperties'.

14   export function definePattern<T>(config: PatternConfig<T>): PatternConfig
                                                            ~

  styled-system/types/global.d.ts:14:33
    14   export function definePattern<T>(config: PatternConfig<T>): PatternConfig
                                       ~
    This type parameter might need an `extends PatternProperties` constraint.

```

no


packages/generator/src/artifacts/types/main.ts
```diff
@@ -14,7 +14,7 @@ export const generateTypesEntry = () => ({
       export function defineGlobalStyles(definition: GlobalStyleObject): GlobalStyleObject
       export function defineTextStyles(definition: CompositionStyles['textStyles']): CompositionStyles['textStyles']
       export function defineLayerStyles(definition: CompositionStyles['layerStyles']): CompositionStyles['layerStyles']
-      export function definePattern<T>(config: PatternConfig<T>): PatternConfig
+      export function definePattern<T extends PatternConfig>(config: PatternConfig<T>): PatternConfig
```
kidjp85, line 17: I think it should be  `  export function definePattern<T extends PatternProperties>(config: PatternConfig<T>): PatternConfig
` instead.

PR #953, opened 2023-07-05
feat(types): make RecipeVariantProps only optional for variants missing defaults

fredericoo:
Closes # 

## 📝 Description

This PR severely modifies types to achieve a specific outcome with `RecipeVariantProps`. It’s a bit of a challenge, but I believe it adds a significant bit of DX to the utility.

`RecipeVariantProps` today returns all props as optional. Regardless of whether they have defaults or not.

## 🚀 New behavior

- Added default props generics to all the relevant utility type functions
- The resulting type of `RecipeVariantProps` now behaves as follows:
  - if the variant _is present_ in recipes `defaultVariants`, then it is optional
  - otherwise, it is required.
 This is not achievable without passing the generic for `defaultVariants` to it.

CleanShot 2023-07-05 at 15 46 07@2x
Example where `size` is optional, but `intent` isn’t, as there is no default.

### DIFF
--- a/.changeset/gold-tips-suffer.md
+++ b/.changeset/gold-tips-suffer.md
@@ -1,0 +1,5 @@
+---
+'@pandacss/types': patch
+---
+
+Utility type `RecipeVariantProps` now respects recipes `defaultVariant`s, making variants that have defaults optional, whilst keeping the required ones required.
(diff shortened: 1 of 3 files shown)


segunadebayo:
Thanks for working on this @fredericoo

After thinking about this and discussing it with @astahmer, we think it's best to move to the userland, even though we see its benefits.

An alternative is to use a `RequiredBy` type utility to pick out the required fields.

```ts
export type RequiredBy<T, K extends keyof T> = Partial<Omit<T, K>> & Required<Pick<T, K>>
```

We'll definitely reconsider this in the future; for now, please use the type utility.

Thanks for understanding.

PR #996, opened 2023-07-09
fix(cli): parse no-gitignore flag properly on `panda init` command

custardcream98:
## 📝 Description

I have made some improvements to the `panda init` command. Previously, the command was not properly parsing the `--no-gitignore` flag. I have fixed this issue in my changes.

When running `pnpm panda init` command, the `setupGitIgnore` function was not being executed, even if the `--no-gitignore` flag was not used.

## 🚀 New behavior

When running `pnpm panda init`, the `setupGitIgnore` function will be executed as intended.

Nope

## 📝 Additional Information

I believe setting up the `.gitignore` file should be the default behavior. Why do we need the `--no-gitignore` option?


segunadebayo:
Thanks for working on this PR @custardcream98.

Based on the CLI framework (cac) docs, using the `no-*` is a negation flag:
https://github.com/cacjs/cac#negated-options

This works correctly on my end. Not sure about the issues you faced.

PR #1018, opened 2023-07-11
fix(conditions): fix some typos in conditons

clementprevot:
## 📝 Description

I noticed a typo in the `highContrast` condition.
And I also noticed some leading spaces for the `dark` and `light` conditions

## ⛳️ Current behavior (updates)

The `highContrast` condition is named `highConstrast`

## 🚀 New behavior

The `highContrast` condition is properly named

## 💣 This is a breaking change

It might be a breaking change because some people might have worked around the issue and are using `highConstrast`

## 📝 Additional Information

Regarding the leading spaces in the `dark` and `light` conditions, it might be on purpose, so don't hesitate to tell me and I will revert this change

PR #1019, opened 2023-07-11
feat(conditions): add negative conditions for all basic conditions

clementprevot:
## 📝 Description

We have a lot of conditions shortcuts to use in the CSS (like `_hover`) but as soon as you want to define the opposite condition, you have to manually rewrite those conditions.

So I figured that it would be a nice thing to also have the negative conditions of the ones provided by PandaCSS.

When you want to use opposite conditions of the one that have shortcuts, you have to manually rewrite those conditions

## 🚀 New behavior

All basic conditions provided by the base preset now have there negative counterpart.

### DIFF
--- a/packages/preset-base/src/conditions.ts
+++ b/packages/preset-base/src/conditions.ts
@@ -1,4 +1,4 @@
-export const conditions = {
+const positiveConditions = {
   hover: '&:is(:hover, [data-hover])',
   focus: '&:is(:focus, [data-focus])',
   focusWithin: '&:focus-within',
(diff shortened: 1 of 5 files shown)


clementprevot:
Why is this PR embarking changes I haven't made on this branch (those regarding the `highConstrast` and leading space typos, cf #1018) 🤔 
I do take care of reset the `feat/negativeConditions` branch on `upstream/main`

segunadebayo:
Hi @clementprevot,

Thanks for working on this. We don’t intend to have the negative conditions in the default config.

Please consider adding it your own panda config or publish a community preset for others to use as well.

We’ll consider this later down the line if it’s really needed.

Thanks for understanding.

Issue #868, opened 2023-06-29
Using TypeScript paths/baseUrl to alias imports makes patterns not work


osdiab:
### Description

In my NextJS app, I use TypeScript's `paths` configuration so that `src/*` files are referred to as `@/*`. I want to handle all local imports with absolute paths this way, but Panda patterns seem to break when referring to its imports like that. Some simple styles provided by `css()` work, but when I try to use the `hstack()` pattern for example, it inserts the classes but the necessary CSS isn't included in the stylesheet.

this happens both when using a paths config, but also just when using baseUrl - absolute imports don’t seem to lead to the corresponding CSS to be included when importing patterns.

### Link to Reproduction

https://github.com/osdiab/nextjs-panda-css-paths-repro/

### Steps to reproduce

1. set up a NextJS app
2. set up Panda
3. set up a value in `paths` in `tsconfig.json` that points to `styled-system` directory
4. try using `hstack()` importing there
5. it doesn't work
6. switch it to a relative import
7. now it works

### JS Framework

NextJS TS

### Panda CSS Version

0.5.0

### Browser

Chrome

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Realized it here: https://github.com/chakra-ui/panda/discussions/867#discussioncomment-6311287
Wrong guess at the cause here: https://github.com/chakra-ui/panda/discussions/847

osdiab:
@astahmer @segunadebayo i tried updating the version of panda in the above example and that doesn't fix it. It looks like #834 was included in the last push so is it safe to say that the fix didn't actually fix the repro case? Thanks

astahmer:
no, that #834 actually was kinda related but didn't fix your case (and the current last public version published, 0.6 doesn't contain those changes anyway) 

on the other hand the PR #868 does fix your issue, thank you for the minimal repro

PR #1039, opened 2023-07-16
Fix/868 ts alias

astahmer:
Closes # https://github.com/chakra-ui/panda/issues/868

## 📝 Description

fix(generator): return import map as array
so that it can be joined later on using the specific runtime that can better handle it for the different OSs etc
this should be related https://github.com/chakra-ui/panda/discussions/980#discussioncomment-6387293

refactor(config): move the tsOptions (baseUrl+pathMappings)
earlier in the pipeline so it can be shared, from the builder (which is postcss-only) to the loadConfigAndCreateContext entrypoint function

fix(config): also pass compilerOptions to resolveModuleName

fix(parser): handle TS path mappings in import regexes as well
otherwise patterns etc wouldnt be recognized
minimal repro https://github.com/chakra-ui/panda/issues/868

refactor(node): rm lil-fp from createContext
cause of TS dts issues

no

### DIFF
--- a/packages/generator/src/generator.ts
+++ b/packages/generator/src/generator.ts
@@ -18,10 +18,10 @@ const defaults = (conf: ConfigResultWithHooks): ConfigResultWithHooks => ({
 })
 
 const getImportMap = (outdir: string) => ({
-  css: `${outdir}/css`,
-  recipe: `${outdir}/recipes`,
-  pattern: `${outdir}/patterns`,
-  jsx: `${outdir}/jsx`,
+  css: [outdir, 'css'],
+  recipe: [outdir, 'recipes'],
+  pattern: [outdir, 'patterns'],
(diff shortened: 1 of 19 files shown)


packages/generator/src/generator.ts
```diff
@@ -32,7 +32,7 @@ export const createGenerator = (conf: ConfigResultWithHooks) => {
   const baseUrl = compilerOptions.baseUrl ?? ''
 
   const cwd = conf.config.cwd
-  const relativeBaseUrl = baseUrl ? baseUrl.replace(cwd, '').slice(1) + '/' : cwd
+  const relativeBaseUrl = baseUrl !== cwd ? baseUrl.replace(cwd, '').slice(1) + '/' : cwd
```
astahmer, line 35: let's also refactor that '/' to something OS agnostic, resolved by path later on

PR #1066, opened 2023-07-19
Refactor generated recipe js code

anubra266:
Refactor generated recipe code, so there are no clashes when several recipes are used in same local scope.

### DIFF
--- a/packages/generator/__tests__/generate-recipe.test.ts
+++ b/packages/generator/__tests__/generate-recipe.test.ts
@@ -78,24 +78,20 @@ describe('generate recipes', () => {
 
       const textStyleFn = createRecipe('textStyle', {}, [])
 
-      const variantKeys = [
-        \\"size\\"
-      ]
-
-      function splitVariantProps(props) {
-        return splitProps(props, variantKeys)
-      }
-
       export const textStyle = Object.assign(textStyleFn, {
         __recipe__: true,
-        variantKeys,
+        variantKeys: [
+        \\"size\\"
+      ],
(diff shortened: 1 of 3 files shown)


packages/generator/__tests__/generate-recipe.test.ts
```diff
@@ -78,24 +78,20 @@ describe('generate recipes', () => {
...
+      ],
         variantMap: {
         \\"size\\": [
           \\"h1\\",
           \\"h2\\"
         ]
       },
-        splitVariantProps,
+        splitVariantProps(props) {
+          return splitProps(props, variantKeys)
```
segunadebayo, line 93: There's no reference to variantKeys here and this line will throw an error.

You can consider inlining the code for `variantKeys` there as well

packages/generator/src/artifacts/js/recipe.ts
```diff
@@ -76,17 +76,13 @@ export function generateRecipes(ctx: Context) {
...
-        }
-        
         export const ${name} = Object.assign(${name}Fn, {
           __recipe__: true,
-          variantKeys,
+          variantKeys: ${stringify(Object.keys(variantKeyMap))},
           variantMap: ${stringify(variantKeyMap)},
-          splitVariantProps,
+          splitVariantProps(props) {
+            return splitProps(props, variantKeys)
```
segunadebayo, line 84:
```suggestion
            return splitProps(props, ${stringify(Object.keys(variantKeyMap))})
```

PR #1067, opened 2023-07-20
playground ux improvements

anubra266:
## Improvements

[x] Implement recipes
[x] replace @next/font with next/font 
[x] Improve token colours
[x] Use fira code font
[x] Add ligatures

CleanShot 2023-07-20 at 02 14 08@2x


segunadebayo:
**Quick feedback:**
- Not all editors use the Fira Code font
- Let's add font aliasing or reduce the font-weight coz because the Fira Code font looks too bold and not as crisp

CleanShot 2023-07-20 at 11 53 05@2x

Issue #1057, opened 2023-07-18
Missing styles when using config recipes and Storybook


rcoedo:
### Description

Panda loses track of used styles when writing [Storybook stories using args](https://storybook.js.org/docs/react/writing-stories/introduction#using-args).

The bug can be reproduced [using this repository](https://github.com/rcoedo/pandacss-pnpm-turbo). The steps to reproduce it are detailed in the README.md.

Thanks for the hard work!

### Link to Reproduction

https://github.com/rcoedo/pandacss-pnpm-turbo

### Steps to reproduce

1. Install the dependencies
2. Run `pnpm dev` to start the `front` and `workshop` applications.
3. Open [workshop](http://localhost:6006/) and [front](http://localhost:5173/) in your browser.
4. Notice that the `Edgy` story is missing its styles.
5. Edit `packages/design-system-react/src/button/button.stories.ts`, comment out the Edgy args story, uncomment the Edgy
   render story.
6. Restart the servers.
7. Notice that `Edgy` is now correctly styled.

### JS Framework

React (TS)

### Panda CSS Version

0.7.0

### Browser

Any

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

astahmer:
I'm not sure if this is what you expect, but this:

https://github.com/rcoedo/pandacss-pnpm-turbo/blob/main/packages/design-system-react/src/button/button.stories.tsx#L29-L36

```ts
export const Edgy: Story = {
	args: {
		visual: "edgy",
		shape: "square",
		size: "sm",
		children: "edgy button",
	},
};
```

can't be extracted by panda as we have no way to know that this object is related to styling

---
on the other hand, this should be just fine:

```tsx
export const FunkyRenderStory: Story = {
	render: () => (
		<Button shape="square" visual="funky" size="lg">
			I'm a button!
		</Button>
	),
};
```

---

if you want to keep using an object, then you could do something like:

```ts
export const Edgy: Story = {
	args: {
		className: css({
			visual: "edgy",
			shape: "square",
			size: "sm",
			children: "edgy button",
		})
	},
}

segunadebayo:
Closing this as the new `.raw` syntax will help hint the extractor.

More details here: https://github.com/chakra-ui/panda/pull/1069

ShakurOo:
What is `button` in your exemple ? A component ? A recipe ? 

`button.raw` because I can't use it with recipe : `Property 'raw' does not exist on type 'RecipeConfig'.`

PR #1069, opened 2023-07-20
feat(parser): `{fn}.raw` syntax

astahmer:
## 📝 Description

@segunadebayo had this awesome idea:

This will allow for a super flexible usage and extraction.

```tsx
<Button rootProps={css.raw({ bg: "red.400" })} />

// recipe in storybook
export const Funky: Story = {
	args: button.raw({
		visual: "funky",
		shape: "circle",
		size: "sm",
	}),
};

// mixed with pattern
const stackProps = {
  sm: stack.raw({ direction: "column" }),
  md: stack.raw({ direction: "row" })
}

stack(stackProps[props.size]))
```

The .raw will be an identity function. Only serves as a marker for extractor

no

## 📝 Additional Information

related to https://github.com/chakra-ui/panda/issues/1057

### DIFF
--- a/packages/parser/src/parser.ts
+++ b/packages/parser/src/parser.ts
@@ -28,6 +28,7 @@ const isNodeRecipe = (node: ParserNodeOptions): node is ParserRecipeNode => node
 
 const cvaProps = ['compoundVariants', 'defaultVariants', 'variants', 'base']
 const isCva = (map: BoxNodeMap['value']) => cvaProps.some((prop) => map.has(prop))
+const isRawFn = (name: string) => Boolean(name.endsWith('.raw'))
 
 export type ParserOptions = {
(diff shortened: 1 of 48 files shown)


packages/parser/src/parser.ts
```diff
@@ -220,7 +221,7 @@ export function createParser(options: ParserOptions) {
 
     const matchFn = memo((fnName: string) => {
       if (recipes.has(fnName) || patterns.has(fnName)) return true
-      if (fnName === cvaAlias || fnName === cssAlias || isFactory(fnName)) return true
+      if (fnName === cvaAlias || fnName === cssAlias || isRawFn(fnName) || isFactory(fnName)) return true
```
segunadebayo, line 224: This logic seems error-prune. It seems to match anything that ends with `.raw` and could be invalid sometimes.

For example:
```js
const x = String.raw``
```

PR #1073, opened 2023-07-20
refactor(cli): interactive mode with clack

astahmer:
## 📝 Description

Screenshot 2023-07-21 at 00 18 00

no, opt-in with -i flag

### DIFF
--- a/.changeset/cyan-kids-cover.md
+++ b/.changeset/cyan-kids-cover.md
@@ -1,0 +1,5 @@
+---
+'@pandacss/dev': minor
+---
+
+Add the CLI interactive mode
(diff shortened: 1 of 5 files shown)


anubra266:
CleanShot 2023-07-30 at 01 20 35@2x

This is only for CLI, so I'm thinking we could just show it only when user does not choose postcss;

segunadebayo:
What’s left to get this in? @astahmer

I would say let’s set interactive flag to false by default to keep the current workflow, and make it an opt in.

Other than that, it looks good to me.

astahmer:
I think other than a ascii logo that we discussed, it's ready
got it, will rebase, revert the breaking change and make it opt-in instead

astahmer:
@anubra266 good points but let's do that in another PR, this is a bit unrelated

anubra266:
May we document the interactive flag.
https://panda-css.com/docs/references/cli#flags
https://panda-css.com/docs/installation/cli

PR #1086, opened 2023-07-23
Playground Improvements

anubra266:
- [x] Add responsive mode
CleanShot 2023-07-24 at 09 35 20@2x
CleanShot 2023-07-24 at 00 24 42@2x

- [x] Improve Errors
CleanShot 2023-07-24 at 11 40 24@2x
CleanShot 2023-07-24 at 11 41 41@2x

- [x] Fix segmented control checked state ui
- [x] Fix color mode switch size shift
- [x] Automatically copy URI after saving playground, and show toast feedback


playground/panda.config.ts
```diff
@@ -39,12 +39,27 @@ export default defineConfig({
...
         },
       },
       recipes,
     },
   },
+  staticCss: {
+    css: [
+      {
+        properties: {
+          cursor: ['ew-resize', 'ns-resize', 'nwse-resize', 'nesw-resize'],
```
astahmer, line 58: not sure this is needed since they are used in the `playground/src/components/Preview/responsive-border.ts` file

astahmer (approved):
insane work ! the responsive view is a really cool feature, well played

playground/panda.config.ts
anubra266: @astahmer It's for the `playground/src/hooks/useResponsiveView.ts` file.
When we're resizing, we change the body cursor. Show it shows that we're resizing.

CleanShot 2023-07-24 at 09 45 54@2x

astahmer: Screenshot 2023-07-24 at 10 49 30
yes but they are used here so panda should be able to extract them already, no ?

anubra266: You're right @astahmer

Issue #1082, opened 2023-07-22
Patterns not working properly when being extended inside a component


omarkhatibco:
### Description

Maybe the title is not the best but let's have some example here.

so I used the `Circle` pattern to make a skeleton component that accept the same props expect it has extra bgColor

now the code of the component look like this 

```ts
import { forwardRef } from 'react';
import { Circle, CircleProps } from '../styled-system/jsx';

export const SkeletonCircle = forwardRef<HTMLDivElement, CircleProps>(
  (props, ref) => {
    return <Circle ref={ref} bgColor='surface.container.highest' {...props} />;
  },
);
```

now when I use the component inside my code like this 

```ts
  <SkeletonCircle size='100px' />
```

now the styles for the this component is unfortunately is not generated specifically `width` & `height`

### Link to Reproduction

[...]

PR #1097, opened 2023-07-24
refactor(core): patterns/recipes + add pattern.jsx

astahmer:
related to: https://github.com/chakra-ui/panda/issues/1082#issuecomment-1647554213

## 📝 Description

refactor(core): patterns/recipes + add pattern.jsx

yes

changed `recipe.name` to -> `recipe.className`
changed `pattern.jsx` to -> `pattern.jsxName`,  added `pattern.jsx` (just like `recipe.jsx`)


thetiby:
@astahmer Thank you for the fixes, improvements and features! 

This is something most likely already on your list, but just to make sure: will pattern have both `jsx` and `jsxName` as options? Or will it have only `jsxName` which was initially named `jsx`, but it was just renamed?

astahmer:
@thetiby 

yes patterns will have both `jsx` (same as recipe.jsx, list of custom components using that pattern) and jsxName (exported component name)

packages/core/src/recipes.ts
```diff
@@ -73,9 +77,14 @@ export class Recipes {
       match,
       config: recipe,
       splitProps: (props) => {
-        return splitProps(props, variantKeys) as [Dict, Dict]
+        const [recipeProps, styleProps] = splitProps(props, variantKeys) as [Dict, Dict]
+        // omit other's recipe variant keys from the style props
+        return [recipeProps, omit(styleProps, Array.from(sharedState.variantKeys))]
```
segunadebayo, line 82: I’m not sure I understand the reasoning behind this. Feels like this might have some unintended effects

PR #1116, opened 2023-07-27
Preset: update conditions in base preset

anubra266:
update conditions in base preset


playground/src/hooks/usePanda.ts
```diff
@@ -8,17 +8,19 @@ import { createHooks } from 'hookable'
 import { useMemo } from 'react'
 import { getResolvedConfig } from '@/src/lib/resolve-config'
 import { merge } from 'merge-anything'
+import { Config, Preset } from '@pandacss/types'
+import { RecipeRule } from '@pandacss/types/dist/static-css'
```
segunadebayo, line 12: This import feels weird. Could you update it?

PR #1124, opened 2023-07-28
Playground: add JSX support

anubra266:
add JSX support to playground

CleanShot 2023-07-28 at 17 53 24@2x


packages/is-valid-prop/scripts/prepare.ts
```diff
@@ -30,18 +30,16 @@ writeFileSync(
...
-  function memo<T>(fn: (value: string) => T): (value: string) => T {
+  function memoize<T>(fn: (value: string) => T): (value: string) => T {
     const cache = Object.create(null)
     return (arg: string) => {
       if (cache[arg] === undefined) cache[arg] = fn(arg)
       return cache[arg]
     }
   }
 
-  const selectorRegex = /&|@/
```
astahmer, line 41: re-using a single Regex instance is actually better for perf reasons
otherwise it would be recreated on each `isCssProperty` call, and since it's being called quite a lot, it matters

packages/is-valid-prop/src/index.ts
```diff
@@ -541,18 +541,16 @@ const allCssProperties = [
 
 const properties = new Map(allCssProperties.map((prop) => [prop, true]))
 
-function memo<T>(fn: (value: string) => T): (value: string) => T {
+function memoize<T>(fn: (value: string) => T): (value: string) => T {
```
astahmer, line 544: I'd rather keep it named `memo` as I'm relying on this name
by removing this memo function declaration and importing the one from `styled-system/helpers.mjs` instead
https://github.com/chakra-ui/panda/pull/1115/files#diff-6f3b94515acc3625211140aac67d30d11205932fd8414b8ffb97b6a2a5c8c997R32

PR #1144, opened 2023-07-31
feat(preset-base): add opacity utility

anubra266:
Add the `opacity` utility to the base preset.


Msordet:
Thank you @anubra266 ! 🔥

PR #1161, opened 2023-08-02
feat: allow customizing layers

### LINKED ISSUE #1164: Panda css layers conflict with tailwind layers
### Description

Hi! Thank you for bringing Panda to life, I've loved the dx so far 

I'm using Panda to build a component library, when it comes to shipping it as a library, I found an issue with my current apps stack. My apps use Tailwind, and now I want to use both my component library and Tailwind in the same app (I'm talking about a Qwik app). But I found that Tailwind and Panda use the same name for some CSS layers, for example `base` and `utilities`

### Problem Statement/Justification

I'm getting the following error.

```
'@layer base' is used but no matching '@tailwind base' directive is present.
```

And this is a simplified version of my main css

```css
@import "@goevery/material-qwik/style"; /* My component library */
@import "./theme/tokens.css"; /* Some CSS var definitions */

@layer reset, base, tokens, recipes, utilities; /* Panda layers */

/* Tailwind config */
@tailwind base;
@tailwind components;
@tailwind utilities;
```

I already tried some solutions, like this import before importing my library CSS `@import 'tailwindcss/base';` but none of them work except for manually changing the name of the generated layers

### Proposed Solution or API

I think a good approach to maximize compatibility with other stacks is to add a prefix to the layers name, something like you did with class names prefixes.

### Alternatives

_No response_

### Additional Information

_No response_

anubra266:
Allow customizing layers Closes #1164
- [x] Builder
- [x] AtomicRule
- [x] Recipes
- [x] Stylesheet

CleanShot 2023-08-03 at 02 34 08@2x


packages/generator/src/artifacts/css/flat-css.ts
```diff
@@ -10,7 +10,9 @@ export const generateFlattenedCss = (ctx: Context) => (options: { files: string[
   const { theme: { keyframes } = {}, preflight, minify, staticCss } = ctx.config
 
   const unresolved = [
-    '@layer reset, base, tokens, recipes, utilities;',
+    `@layer ${Object.values(ctx.layers)
+      .map((l, i, arr) => (i === arr.length - 1 ? l : `${l}, `))
+      .join('')};`,
```
astahmer, line 15: I'm not sure about this, when running this code:

```ts
const layers = {  base: "base", tokens: "tokens", recipes: "recipes", reset: "reset", utilities: "utilities", other: "other" }

Object.values(layers).map((l, i, arr) => (i === arr.length - 1 ? l : `${l}, `))
      .join('')
```

I get this result, which is in the wrong order : `'base, tokens, recipes, reset, utilities, other'`

---
something like this looks a bit better

```ts
const layers = {  base: "base_suffixed", tokens: "tokens", recipes: "recipes", reset: "reset", utilities: "panda_utilities", other: "other" }

const layersStr = '@layer reset, base, tokens, recipes, utilities;'
const updated = Object.keys(layers).reduce((acc, name) => acc.replace(name, layers[name]), layersStr)

console.log(updated)
// `@layer reset, base_suffixed, tokens, recipes, panda_utilities;`
```

astahmer (approved):

juanpmarin:
Is this getting merged soon? :(

PR #1195, opened 2023-08-10
feat(generator): better cx

astahmer:
Closes https://github.com/chakra-ui/panda/discussions/874 https://github.com/chakra-ui/panda/discussions/931 https://github.com/chakra-ui/panda/discussions/1131

## 📝 Description

Make the `cx` smarter by merging and deduplicating the styles passed in

Example:

```tsx
<h1 className={cx(css({ mx: '3', paddingTop: '4' }), css({ mx: '10', pt: '6' }))}>Will result in "mx_10 pt_6"</h1>
```

kinda ? in a good way


astahmer:
currently cannot work due to RSC. 🫠
cant serialize `toString` since it's a function, we have to find another way to do that I guess

packages/studio/styled-system/css/css.mjs
```diff
@@ -28,15 +28,22 @@ const context = {
...
   },
 }
 
-export const css = createCss(context)
+const cssFn = createCss(context)
+export const cssCache = new Map()
+export const css = (styles) => {
+  const classNames = cssFn(styles)
+  cssCache.set(classNames, styles)
+  console.log({ classNames, styles })
```
segunadebayo, line 44: Looks good. Can we remove this log line?

packages/generator/src/artifacts/js/css-fn.ts
```diff
@@ -100,17 +100,24 @@ export function generateCssFn(ctx: Context) {
...
+            : `(key, value) => ({ className: \`$\{classNameByProp.get(key) || hypenateProperty(key)}${separator}$\{withoutSpace(value)}\` })`
         },
         ${utility.hasShorthand ? 'hasShorthand: true,' : ''}
         resolveShorthand: ${utility.hasShorthand ? 'resolveShorthand' : 'prop => prop'},
       }
     }
 
-    export const css = createCss(context)
+    const cssFn = createCss(context)
+    export const cssCache = new Map()
```
segunadebayo, line 114: We should consider using an LRU Cache here to avoid racking up too much memory. 

Creating a map cache that stores up to a certain size (say 500 entries) and resets afterward might help.

packages/studio/styled-system/css/css.mjs
astahmer: done !

theMasix:
@segunadebayo https://github.com/chakra-ui/panda/pull/2056#issuecomment-1902676458

astahmer:
> @segunadebayo [#2056 (comment)](https://github.com/chakra-ui/panda/pull/2056#issuecomment-1902676458)

this has been discarded in favor of css.raw a while ago 
https://github.com/chakra-ui/panda/blob/main/CHANGELOG.md#changed-15
https://panda-css.com/docs/concepts/merging-styles

theMasix:
@astahmer 
Thanks for your info.
What if we want to merge styles of some hierarchical components?

Suppose we have `IconButton` > `Button` > `ButtonBase` components that each of them is a wrapper around another

We want IconButton's styles to have the highest priorities.

Keep in mind that we pass `IconButton`'s styles to Button and Button's to `ButtonBase` using plain JSX `className` property.

We do not pass CSS object from the wrapper component. We only pass result of `cva`

astahmer:
then the `cx` approach makes sense yes

theMasix:
@astahmer It doesn't work when you enable hashing.

The #2056 solves the issue.

PR #1204, opened 2023-08-11
feat: adds support for generating staticCss for Panda patterns

thedanchez:
Closes # 

## 📝 Description

This PR adds the new capability to generate static CSS for all the out of the box Panda patterns (`hstack`, `vstack`, etc.)

Current behavior does not support this feature.

## 🚀 New behavior

Static CSS can now be generated for the default patterns via the new `patterns` key under the `staticCss` config field.

No breaking changes

## 📝 Additional Information

This PR would allow component library authors that are pursuing the static CSS path to export out the default Panda patterns alongside their custom components which then would allow said component-library users to leverage the JSX patterns like `HStack` and `VStack` in their applications without the need to use Panda in their apps.

### DIFF
--- a/packages/core/__tests__/static-css.test.ts
+++ b/packages/core/__tests__/static-css.test.ts
@@ -28,6 +28,14 @@ const ctx = {
     }
     return values[property] ?? []
   },
+  getPatternKeys: (property: string) => {
+    const values: Record<string, any> = {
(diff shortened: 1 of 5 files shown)


anubra266:
Thanks for working on this @thedanchez feel free to reopen when it's ready

PR #1211, opened 2023-08-13
[generator] narrow down the `!important` modifier type implementation

isBatak:
## 📝 Description

I think we should only allow strings with the trailing `!`.

Here is the TS playground demonstration https://www.typescriptlang.org/play?#code/C4TwDgpgBAwg9gGzgJwAoEMEWMaBeKAImTgGcJCoAfIsASwDsBrSmgClOGUYHMoAyKAG8AvgEoA3ACgA9DKgKAegH4pU0JFiIUGLDggB1OsAAWASQC2YFMHQNgUAsTIVqtRizcADACRDO3Aw8IgCEXrLySqpSAMZwDJxQAK4J6ABmEAD6EAAe6FZYAIwAXFpIaJjYuI5QAOQk5LXScQkOKaTpWbn5YFgATKXw5bpV+HUNECFNsfGJ7Z3ZeQUQAMyD2hV61QT1Lj5NUHJQmKRwxwhIAO4QACYzrcmpGYs9WAAs68OV+jW1KUwMOCXBgHI4nM6YK63NQtRIdZ7dZYlMo6b64IymSzWZC2ey-Ca1e5whaI3oQAYozajDHmKw2OwOHYTKZEhzwrpLMlrSkjfQ0rH0vFMva1Q7yQEOSFA6GwtkkznvT6oraGYy07G4xl1f6A4FTMVQUx0UiPNI2FLoXAIEBQS4oJikAA0UAARkkHHQ0rboDczkaTb4hENlaNQl5DSYIAxju64ABaUhJHg8CCcOjxW10C5QFq2RiuuCmKBeCbhuw3YvM8JAA

Current implementation allows any arbitrary string to be passed as a styled prop.

## 🚀 New behavior

It will throw an TS error if string contain anything else except trailing `!`, for example `colorPalette: "rose$"`

No

### DIFF
--- a/packages/generator/__tests__/generate-prop-types.test.ts
+++ b/packages/generator/__tests__/generate-prop-types.test.ts
@@ -307,9 +307,9 @@ describe('generate property types', () => {
       }
 
       export type PropertyValue<T extends string> = T extends keyof PropertyTypes
-        ? ConditionalValue<PropertyTypes[T] | CssValue<T> | (string & {})>
+        ? ConditionalValue<PropertyTypes[T] | CssValue<T> | \`\${string}!\`>
         : T extends keyof CssProperties
-        ? ConditionalValue<CssProperties[T] | (string & {})>
+        ? ConditionalValue<CssProperties[T] | \`\${string}!\`>
         : ConditionalValue<string | number>"
     `)
   })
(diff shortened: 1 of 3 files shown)


isBatak:
After some more testing I realised that It's not working properly.

segunadebayo:
Let’s close this out for now. 

When you have a working solution, feel free to open it again.

PR #1237, opened 2023-08-19
fix: export types from @pandacss/dev

### LINKED ISSUE #1122: @pandacss/dev should export types
### Description

`@pandacss/dev` should export types for every function it exports. e.g. `Preset`.

### Problem Statement/Justification

When creating a preset, Typescript will complain because it can't resolve the inferred type. 

```ts
export const preset = definePreset({});
```

```
The inferred type of 'preset' cannot be named without a reference to
'.pnpm/@pandacss+types@0.8.0/node_modules/@pandacss/types'. 
This is likely not portable. A type annotation is necessary. ts(2742)
```
https://typescript.tv/errors/#TS2742

### Proposed Solution or API

The package should export all the types from `@pandacss/types`.

### Alternatives

You can manually add the type like so:

```ts
type Preset = ReturnType<typeof definePreset>;

export const preset: Preset = definePreset({});
```

### Additional Information

I'm using PNPM with NX.

anubra266:
Export all types from `dev` package
Closes #1122


packages/cli/src/index.ts
```diff
@@ -15,6 +15,8 @@ import type {
   Tokens,
 } from '@pandacss/types'
 
+export type * from '@pandacss/types'
```
segunadebayo, line 18: We need to pick out the types explicitly to avoid breaking changes in the future.

The things we need to export are mostly related to all the `define*` helper functions (Config, RecipeConfig, Preset, Pattern).

.changeset/fuzzy-lemons-greet.md
```diff
@@ -0,0 +1,6 @@
+---
+'@pandacss/studio': patch
+'@pandacss/dev': patch
+---
+
+Export all types
```
segunadebayo, line 6:
```suggestion
Export types for all `define` helper functions
```

PR #1241, opened 2023-08-20
feat: add css prop to recipes/patterns functions

astahmer:
## 📝 Description

Add `css` prop to recipes and patterns so you don't have to import and use `cx` for that one little edge case.

Before:

```tsx
<button className={cx(button({ variant: 'primary', state: 'focused' }), css({ color: 'yellow' }))}>Click me</button>
```

Now:

```tsx
<button className={button({ variant: 'primary', state: 'focused', css: { color: 'yellow' } })}>Click me</button>
```

no

### DIFF
--- a/.changeset/soft-moons-approve.md
+++ b/.changeset/soft-moons-approve.md
@@ -1,0 +1,17 @@
+---
+'@pandacss/generator': patch
+---
+
+Add `css` prop to recipes and patterns so you don't have to import and use `cx` for that one little edge case.
+
+Before:
+
+```tsx
+<button className={cx(button({ variant: 'primary', state: 'focused' }), css({ color: 'yellow' }))}>Click me</button>
+```
+
+Now:
+
+```tsx
+<button className={button({ variant: 'primary', state: 'focused', css: { color: 'yellow' } })}>Click me</button>
+```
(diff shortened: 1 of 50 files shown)


anubra266:
Can we add the docs too.

segunadebayo:
I'm open to this idea, but at the same time, I'm not convinced we "really" need this in Panda.

That said, this feature might only be helpful in recipes. Patterns already allow overriding styles since they are just sugar syntax on top of the `css` function.

We also need to include the version this was introduced in the docs.

astahmer:
yeah we discussed that with @anubra266 and came to the conclusion that:
- ✅ for recipes, it's a nice convenient property that avoids importing `cx` and `css`
- ❌ enabling it for patterns doesn't really make sense since we can already override styles, except to stay consistent with the APIs 
- ⚠️ we can only have that in config recipe (NOT `cva`) because we can't know when a function is a `cva` one, due to the usual static limitations, unless using `css.raw`


✅ `badge({ status: 'warning', css: css.raw({ background: 'purple.300' }) })`
❌ `badge({ status: 'warning', css: { background: 'purple.300' } })`

which: 
- defeats the original purpose of not having to import `css`, only the `cx` part is skipped
- introduces inconsistent API difference between `cva` / `config recipe` runtime usage

---

Conclusion: cancel this PR ? @anubra266 @segunadebayo

segunadebayo:
Closing based on mutual agreement

PR #1243, opened 2023-08-20
chore: show panda's version in playground

anubra266:
Show latest version in playground
 
CleanShot 2023-08-21 at 10 44 26@2x


playground/src/components/Toolbar.tsx
```diff
@@ -1,10 +1,19 @@
 import { HStack } from '@/styled-system/jsx'
 import { PropsWithChildren } from 'react'
 import { Logo } from './Logo'
+import { css } from '@/styled-system/css'
+import { version } from 'package.json'
```
segunadebayo, line 5: This is pointing to version of the playground, not panda's version.

We need to update it to something like @pandacss/dev/package.json

PR #1263, opened 2023-08-24
Fix `AnimationName` type

anubra266:
Fix issue where `AnimationName` type was generated wrongly if keyframes were not resolved

### DIFF
--- a/packages/generator/src/artifacts/types/token-types.ts
+++ b/packages/generator/src/artifacts/types/token-types.ts
@@ -52,9 +52,9 @@ export function generateTokenTypes(ctx: Context) {
       result.add(`\t\t${key}: ${typeName}Token`)
     }
 
-    const keyframes = Object.keys(theme?.keyframes ?? {})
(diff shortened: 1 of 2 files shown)


packages/generator/src/artifacts/types/token-types.ts
```diff
@@ -52,9 +52,9 @@ export function generateTokenTypes(ctx: Context) {
       result.add(`\t\t${key}: ${typeName}Token`)
     }
 
-    const keyframes = Object.keys(theme?.keyframes ?? {})
+    const keyframes = unionType(Object.keys(theme?.keyframes ?? {}))
 
-    set.add(`export type AnimationName = ${unionType(keyframes)}`)
+    if (keyframes) set.add(`export type AnimationName = ${keyframes}`)
```
segunadebayo, line 57:
```suggestion
    if (theme?.keyframes) set.add(`export type AnimationName = ${unionType(Object.keys(theme.keyframes))}`)
```

You can remove L55 with this

PR #1283, opened 2023-08-28
fix: vue parser

### LINKED ISSUE #1281: Panda does not detect styles after nested template
### Description

```vue
<template>
  <div @click="selectRole" :class="rootStyles">
    <div :class="rowStyles">
      <slot name="icon"></slot>
      <div :class="hstack()">
        <p :class="css({ textStyle: 'overline' })">
          <slot name="price"></slot>
        </p>
        <div>
          <template v-if="isSelected">
            <IconRadioSelected />
          </template>
          <template v-else>
            <IconRadio />
          </template>
        </div>
      </div>
    </div>
    <h7 :class="css({ textStyle: 'h7' })"><slot name="heading"></slot></h7>
    <p :class="css({ textStyle: 'text', color: 'grey.70' })">
      <slot name="description"></slot>
    </p>
  </div>
</template>
```
Styles after
```vue
<template v-if="isSelected">
            <IconRadioSelected />
          </template>
          <template v-else>
            <IconRadio />
          </template>
```
 are omitted.

### Link to Reproduction

-

### Steps to reproduce

2. Click on '...'
3. Scroll down to '...'

### JS Framework

Vue

### Panda CSS Version

0.13.0

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

anubra266:
Closes #1281

**The current logic:**
1. uses`String.indexOf` to find the opening and closing `template` tag.
2. Is updating the tag attributes by locating indexes of the props from codestring

**Problems.**
1. Using `String.indexOf` means when there's nested templates, the first closing `template` tag is taken instead of the outermost one.
2. While we're updating the tag attributes, the code string gets mutated, this would mean we'll be searching for wrong locations in the string.

**Solutions:**
1. Instead of finding the template position with String methods or Regex, we can take advantage of the vue compiler which can directly provide us with the template content.
2. Instead of finding tag attributes/props with index locations, we can just find their exact string instead and replace it with the new format.

### DIFF
--- a/packages/parser/src/vue-to-tsx.ts
+++ b/packages/parser/src/vue-to-tsx.ts
@@ -40,7 +40,7 @@ const NodeTypes = {
 export const vueToTsx = (code: string) => {
   try {
     const parsed = parse(code)
-    const fileStr = new MagicString(code)
+    const fileStr = new MagicString(`<template>${parsed.descriptor.template?.content}</template>` ?? '')
 
     const rewriteProp = (prop: BaseElementNode['props'][number]) => {
       if (
@@ -48,7 +48,7 @@ export const vueToTsx = (code: string) => {
         prop.exp?.type === NodeTypes.SIMPLE_EXPRESSION &&
         prop.arg?.type === NodeTypes.SIMPLE_EXPRESSION
       ) {
-        fileStr.update(prop.loc.start.offset, prop.loc.end.offset, `${prop.arg.content}={${prop.exp.content}}`)
+        fileStr.replace(prop.loc.source, `${prop.arg.content}={${prop.exp.content}}`)
       }
     }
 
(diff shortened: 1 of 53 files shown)


packages/parser/src/vue-to-tsx.ts
```diff
@@ -40,15 +40,15 @@ const NodeTypes = {
 export const vueToTsx = (code: string) => {
   try {
     const parsed = parse(code)
-    const fileStr = new MagicString(code)
+    const fileStr = new MagicString(`<template>${parsed.descriptor.template?.content}</template>` ?? '')
```
anubra266, line 43: @segunadebayo @astahmer 
We could do this instead:
```suggestion
const fileStr = new MagicString(parsed.descriptor.template?.content ?? '')
```
Only did the former to match current tests, but I don't suppose we need the template tag anywhere in the Panda process.

astahmer: I think I added it because otherwise ts-morph had a hard time creating the correct AST nodes
you can check if it's fine here
https://ts-ast-viewer.com/

anubra266: It should stay then

segunadebayo:
@anubra266 Can you include a test for the nested template that shows that this works?

anubra266:
Done @segunadebayo

astahmer (approved):

PR #1284, opened 2023-08-29
Allow `.mts` and `.cts` panda config extension

anubra266:
Allow `.mts` and `.cts` panda config extension

### DIFF
--- a/packages/config/src/find-config.ts
+++ b/packages/config/src/find-config.ts
@@ -1,7 +1,7 @@
 import findUp from 'escalade/sync'
 import { resolve } from 'path'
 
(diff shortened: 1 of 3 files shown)


packages/config/src/find-config.ts
```diff
@@ -1,7 +1,7 @@
 import findUp from 'escalade/sync'
 import { resolve } from 'path'
 
-const configs = ['.ts', '.js', '.mjs', '.cjs']
+const configs = ['.ts', '.js', '.mts', '.mjs', '.cjs']
```
segunadebayo, line 4: Did you forget to add cts here?

PR #1287, opened 2023-08-30
feat(recipes): nested @layer _slots inside @layer recipes

astahmer:
## 📝 Description

Add a new `_slots` @layer for slot recipes so that "normal" recipes will have a higher CSS specificity

no

## 📝 Additional Information

original issue 
https://discord.com/channels/1118988919804010566/1120305029056831548/1145979787156926504
https://play.panda-css.com/XJynA6uzUU

### DIFF
--- a/.changeset/rude-wasps-speak.md
+++ b/.changeset/rude-wasps-speak.md
@@ -1,0 +1,5 @@
+---
+'@pandacss/core': patch
+---
+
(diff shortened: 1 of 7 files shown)


segunadebayo:
I managed to implement this in a slightly simpler approach: See https://github.com/chakra-ui/panda/commit/bc3b077d4050a996500fa106b2171f7e128c2662

Thanks for working on this.

PR #1288, opened 2023-08-30
feat(sva): add .raw fn

astahmer:
## 📝 Description

Add `{svaFn}.raw` function to get raw styles and allow reusable components with style overrides, just like with
`{cvaFn}.raw`

no


anubra266:
Thanks for working on this @astahmer. Could you add the docs here: 
https://panda-css.com/docs/concepts/writing-styles#merging-styles
Would be okay after the CVA example, 
**SVA Example:**

PR #1291, opened 2023-08-30
types: make css(...args) allow falsy values

astahmer:
Closes https://github.com/chakra-ui/panda/pull/1264#issuecomment-1698394167

## 📝 Description

`css(...args)` are already filtered to only take objects, this PR makes the typings lenient with falsy values, mostly intended for component props that have optional values like `cssProps?: SystemStyleObject` and then later `css({ ... }, cssProps)`

no

### DIFF
--- a/.changeset/smart-mails-train.md
+++ b/.changeset/smart-mails-train.md
@@ -1,0 +1,8 @@
+---
+'@pandacss/generator': patch
+---
+
+Change the typings for the `css(...args)` function so that you can pass possibly undefined values to it.
+
+This is mostly intended for component props that have optional values like `cssProps?: SystemStyleObject` and would use
+it like `css({ ... }, cssProps)`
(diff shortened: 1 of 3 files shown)


osdiab:
this is definitely an improvement but i'm running into one more usability problem.

right now `css()` can take multiple styles, but `css.raw()` cannot. So if I want to share a single style, that's easy by exporting a `css.raw()` system style object, but if I want to share a complex style comprised of multiple system style objects, I'm left either:

- exporting a `SystemStyleObject[]` (downside: the API of using this style is now not the same as the API of using a single system style object)
- as a convention, exporting all shared styles as `SystemStyleObject[]` even if there is only a single one so that I can have consistent usage (downside: it is not natural to by convention wrap single values in arrays for no obvious reason except in other modules)
- updating usages of exported styles to always go through a `wrapArray()` function that will wrap a single style object in an array if it is not already an array of styles and then splatting that onto `css()` calls (downside: this is a significant amount of boilerplate littering the codebase)

It seems like a natural extension would be to have `css.raw()` also be able to take a list of style objects instead of just a single one and perform the same style merging. Or alternatively, a separate explicit separate composition function that isn't `css()` or `css.raw()`. Thoughts? @astahmer

PR #1298, opened 2023-09-02
Update redwood install instructions to fix Vite resolving issue

freddydumont:
There's a missing step in the installation instructions for Redwood. By default, the Redwood Vite plugin only resolves in `web/src` so we need to add a resolve alias to the config.

[Full writeup](https://freddydumont.com/blog/vite-add-folder) of the issue if you need more info or background. It's a quick read.

I created this PR on GitHub itself so not sure if the formatting is correct.

### DIFF
--- a/website/pages/docs/installation/redwood.mdx
+++ b/website/pages/docs/installation/redwood.mdx
@@ -83,6 +83,23 @@ Update the tsconfig file to include the `styled-system` folder.
 }
 ```
 
+
+Update the Vite config to resolve the `styled-system` folder.
+
+```ts {3-10} filename="web/vite.config.ts
+const viteConfig: UserConfig = {
+  plugins: [redwood()],
+  resolve: {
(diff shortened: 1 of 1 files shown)


anubra266:
@freddydumont Thanks for your PR This is for baseUrl path import though.

```js
import { css } from "styled-sytem/css";
// Instead of 
import { css } from "../styled-sytem/css";
```
So it's not needed. All code examples in the docs use the latter

freddydumont:
@anubra266 I see. I think the issue is that just above, tsconfig is set up for paths imports. 

So the editor respects the config and auto-imports using the path rather than doing relative import. 

At the moment, this particular tutorial suggests using `paths` but this leads to a broken setup. 

Maybe it'd be better to remove the tsconfig part then, or hide the complete `paths` setup behind a foldable block with an explanation?

PR #1309, opened 2023-09-04
feat(utility): allow retrieving the token object

astahmer:
## 📝 Description

The utility transform fn now allow retrieving the token object with the raw value/conditions as currently there's no way
to get it from there.

Example:

Given I have this config.tokens:
`colors: { redish: { value: '#9c1d1d' } }`

And this utility transform:

```ts
{
    utilities: {
        backgroundColor: {
            shorthand: "bgColor",
            className: "bg",
            values: "colors",
            transform(value, args) {
                console.log({ value, args, raw: args.token(args.raw), fromPath: args.token('colors.' + args.raw) })
                return {
                    backgroundColor: `rgba(${convertToRgba(value)}, var(--panda-text-opacity))`,
                };
            },
        },
    },
}
```

the output log would be

```ts
{
  value: 'var(--colors-redish)',
  args: { token: [Function: bound getToken], raw: 'redish' },
  raw: undefined,
  fromPath: 'var(--colors-redish)'
}
```

now it can be retrieved like this:

```
transform(_, args) {
    const token = args.token.raw('colors.' + args.raw)
    return { 
        // ...
    }
}
```

no

### DIFF
--- a/.changeset/polite-shoes-attack.md
+++ b/.changeset/polite-shoes-attack.md
@@ -1,0 +1,7 @@
+---
+'@pandacss/types': patch
+'@pandacss/core': patch
+---
+
+The utility transform fn now allow retrieving the token object with the raw value/conditions as currently there's no way
+to get it from there.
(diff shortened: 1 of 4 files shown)


anubra266:
Thanks for this @astahmer Would be nice to have docs here
https://panda-css.com/docs/customization/utilities

PR #1310, opened 2023-09-04
fix: `token()` CSS syntax error when non-existing

astahmer:
## 📝 Description

Fix issue with the `token()` function in CSS strings that produced CSS syntax error when non-existing token were left
unchanged (due to the `.`)

```css
* {
  color: token(colors.magenta, pink);
}
```

## 🚀 New behavior

```css
* {
  color: token('colors.magenta', pink);
}
```

no

### DIFF
--- a/packages/core/src/plugins/expand-token-fn.ts
+++ b/packages/core/src/plugins/expand-token-fn.ts
@@ -9,7 +9,9 @@ function expandTokenFn(fn?: (str: string) => string | undefined): TransformCallb
       if (decl.value.includes('token(')) {
         const value = decl.value.replace(tokenRegex, (_, token) => {
           const [tokenValue, tokenFallback] = token.split(',').map((s: string) => s.trim())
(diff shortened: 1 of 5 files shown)


packages/core/src/plugins/expand-token-fn.ts
```diff
@@ -9,7 +9,9 @@ function expandTokenFn(fn?: (str: string) => string | undefined): TransformCallb
       if (decl.value.includes('token(')) {
         const value = decl.value.replace(tokenRegex, (_, token) => {
           const [tokenValue, tokenFallback] = token.split(',').map((s: string) => s.trim())
-          const result = [tokenValue, tokenFallback].filter(Boolean).map((s) => fn?.(s) ?? s)
+          const result = [tokenValue, tokenFallback]
+            .filter(Boolean)
+            .map((s) => fn?.(s) ?? (s.includes('.') ? `'${s}'` : s))
```
segunadebayo, line 14: We should consider escaping instead of using `.includes`

PR #1314, opened 2023-09-04
refactor(internal): rm chunks

astahmer:
## 📝 Description

Switching to a in-memory chunk representation to prevent writing/reading chunk files. This also simplifies the current node package API

no


packages/node/src/chunk-engine.ts
```diff
@@ -1,40 +0,0 @@
...
-  getFiles() {
-    const files = fs.existsSync(this.dir) ? fs.readDirSync(this.dir) : []
-    return files.map((file) => fs.readFileSync(path.join(this.dir, file)))
-  },
-  format(file: string) {
-    return path.relative(config.cwd, file).replaceAll(path.sep, '__').replace(path.extname(file), '.css')
-  },
-  getArtifact(file: string, css: string): Artifact {
-    const fileName = this.format(file)
-    const newCss = mergeCss(this.readFile(file), css)
```
segunadebayo, line 25: This line was added previously to avoid janky updates or unstyled flashes when using the CLI. We need to recreate this with the in-memory version.

It should be essentially similar to the Builder used in postcss.

astahmer: I thought so too but I couldn't reproduce this on the panda website locally, is there a way to reproduce this ?

segunadebayo: You can reproduce this by running any of the sandbox examples with the CLI watch flag. No postcss

segunadebayo:
Going to mark this as draft so I can review it later in the week

astahmer:
let's keep that in draft anyway while I continue working on the in memory chunks

astahmer:
closing in favor of https://github.com/chakra-ui/panda/pull/1314

Issue #1315, opened 2023-09-04
CSS not generated with a recipe.jsx regex match


mleralec:
### Description

Props passed to a styled-component generate classes but not CSS when the name of the component matches the regex of a recipe.jsx field. 

This following code works without **recipes** in the configuration. But, if you have a recipe with something like that: `jsx: [/Button.*/]`, the class `bg_pink` doesn't exist in the output CSS.

```tsx
type ButtonProps = HTMLStyledProps<"button">;
const StyledButton = styled("button", { base: { padding: "md" } });

const Button = ({ children, ...props }: ButtonProps) => (
  <StyledButton {...props}>{children}</StyledButton>
);

const TomatoButton = (props: ButtonProps) => (
  <Button backgroundColor="tomato" {...props} />
);

export const Test = () => {
  return (
    <div>
      {/* works: background-color: tomato */}
      <TomatoButton>Button</TomatoButton>
      {/* works: background-color: yellow */}
      <Button backgroundColor="yellow">Button</Button>
      {/* doesn't work, "bg_pink" class is present but no CSS generated  */}
      <TomatoButton backgroundColor="pink">Button</TomatoButton>
    </div>
  );
};

```

### Link to Reproduction

https://play.panda-css.com/O9kEjz3vuP

### Steps to reproduce

Open the reproduction link, remove the **buttonRecipe** of the configuration and it works.

### JS Framework

React (TS)

### Panda CSS Version

0.13.1

### Browser

Google Chrome 116

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

hmiller020414:
I am facing a similar issue with no relation to recipes or regex.

I've built a test example as follows.
```
export const TestBox = ({ cssProps }: { cssProps: SystemStyleObject }) => (
  <div className={css({ padding: 8, bgColor: 'orange', color: 'yellow' }, cssProps)}>
    <p>This is my box</p>
  </div>
);

<TestBox cssProps={{ position: 'absolute', top: '33px' }} />
```

The class `top_33px` is generated in the HTML, but the class `top_33px` does not exist in the output CSS.

astahmer:
that's not related

> @astahmer I am facing a similar issue with no relation to recipes or regex.
> 
> I've built a test example as follows.
> 
> ```
> export const TestBox = ({ cssProps }: { cssProps: SystemStyleObject }) => (
[...]

PR #1316, opened 2023-09-04
fix: JSX components matching using regex leading to missing style props

astahmer:
Closes https://github.com/chakra-ui/panda/issues/1315

## 📝 Description

Fix an issue where some JSX components wouldn't get matched to their corresponding recipes/patterns when using `Regex`
in the `jsx` field of a config, resulting in some style props missing.

no


packages/generator/src/artifacts/css/parser-css.ts
```diff
@@ -29,7 +29,7 @@ export const generateParserCss = (ctx: Context) => (result: ParserResultType) =>
 
       result.jsx.forEach((jsx) => {
         jsx.data.forEach((data) => {
-          sheet.processStyleProps(data)
+          sheet.processStyleProps(filterProps(ctx, data))
```
segunadebayo, line 32: It seems like this filterProps should be happening in the processStyleProps method of the stylesheet.

Can we move it there to avoid the duplicates?

bharat5604 (approved):

mleralec (approved):

PR #1333, opened 2023-09-07
Feat: support functions generating the recipe styles

smoogly:
Closes #1332 

## 📝 Description

Updated ts-evaluator preset to `ECMA` so that parsing
CallExpressions could leverage the evaluator functionality.

This allows generating styles programmatically, as opposed to specifying literals only, e.g.:
```
cva({
  variants: generateVariants()
})
```

At the moment, only the literal styles are parsed, even though `CallExpression`s are matched when attempting to create `BoxNode`s:
https://github.com/chakra-ui/panda/blob/4bc515ea31d9dade49ccdf5d9947cc5220ab38e0/packages/extractor/src/maybe-box-node.ts#L195-L200

This happens because during the evaluation in `ts-evaluator`, `Function` handling is not defined due to limited preset:
https://github.com/chakra-ui/panda/blob/bc3b077d4050a996500fa106b2171f7e128c2662/packages/parser/src/parser.ts#L63

## 🚀 New behavior

Changing the preset to `ECMA` allows evaluating a reasonable set of functions leveraging existing integration with `ts-evaluator`. In turn, this allows generating styles via functions, that are still resolved at build time.

No

### DIFF
--- a/packages/parser/__tests__/cva.test.ts
+++ b/packages/parser/__tests__/cva.test.ts
@@ -312,4 +312,115 @@ describe('ast parser / cva', () => {
       }
     `)
   })
+
+  test('should evaluate variants supplied as a function', () => {
+    const code = `
+    import {cva} from ".panda/css"
+
+    const variants = () => {
+      const spacingTokens = Object.entries({
+          s: 'token(spacing.1)',
+          m: 'token(spacing.2)',
+          l: 'token(spacing.3)',
+      });
+      
+      const spacingProps = {
+          'px': 'paddingX',
+          'py': 'paddingY',
+      };
+      
+      // Generate variants programmatically
+      return Object.entries(spacingProps)
+          .map(([name, styleProp]) => {
+              const variants = spacingTokens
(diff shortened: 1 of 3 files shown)


astahmer:
> Unfortunately, I don't have access to the vercel workflows. Happy to fix the problem with a bit of help — could someone with access let me know what the issue is?

me neither but you can probably reproduce when `build:playground` in the root
also you got a TS error, you can reproduce using the `pnpm typecheck` command in the root

smoogly:
As maintainers you are of course free to make these sort of calls, and you don't even have to explain why. I understand what I'm proposing might not fit the project. However, I don't think the arguments hold.

This only slows down the compiler for those who explicitly opt-into this by using the functions to generate styles. And "not involving runtime code in static analysis" does not seem to be true in the current codebase without any of my involvement:
https://github.com/chakra-ui/panda/blob/4bc515ea31d9dade49ccdf5d9947cc5220ab38e0/packages/extractor/src/maybe-box-node.ts#L195-L200

PR #1339, opened 2023-09-08
(docs) Display changelog on docs site

JuliusGudni:
## 📝 Description

Adds the changelog to docs, instead of redirecting to github

Currently, when "Changelog" is clicked in the, the browser redirects to the github file

## 🚀 New behavior

This PR adds support for fetching and displaying the changelog inline in the docs - with code highlighting, and same style and components as the rest of the docs 

No

### DIFF
--- a/website/pages/docs/overview/_meta.json
+++ b/website/pages/docs/overview/_meta.json
@@ -3,8 +3,5 @@
   "why-panda": "Why Panda?",
   "faq": "FAQs",
   "browser-support": "Browser Support",
-  "changelog": {
-    "title": "Changelog",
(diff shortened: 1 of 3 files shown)


anubra266:
@JuliusGudni The problem with this is the changelog content is too large for the page. 
Something like `/changelog/v0.1.0` would be better, if that's possible

segunadebayo:
Thanks for working on this @JuliusGudni and @anubra266 

We’ll keep the Changelog in the repo for now until we find a better way to manage/trim the changelog based on semver.

Thanks for understanding.

PR #1377, opened 2023-09-14
wip: update package.json

astahmer:


segunadebayo:
With the new importMap PR, I think this can be closed.

The user now has control over codegen outdir and the imports.

PR #1389, opened 2023-09-18
feat(playground): add examples

anubra266:
Add examples in playground
- [x] Add examples
- [x] Link to example
E.g. https://play.panda-css.com/?eg=cva will auto load the cva sample.

CleanShot 2023-09-21 at 15 02 40@2x


segunadebayo:
I think we can improve the look of those examples.
- Try to create an actual button component
- Center the examples on screen
- Let's fix the branch conflicts.

playground/theme/recipes/menu.ts
```diff
@@ -0,0 +1,69 @@
...
+  transition: 'background 0.2s ease',
+  _hover: {
+    bg: { base: 'gray.100', _dark: '#3A3A3AFF' },
+  },
+  _highlighted: {
+    bg: { base: 'gray.100', _dark: '#3A3A3AFF' },
+  },
+} as const
+
+export const menu = defineRecipe({
```
astahmer, line 28: wouldnt it make more sense to use defineSlotRecipes and ditch the defineParts ?

anubra266: @astahmer I'm not using slot recipes so I can apply the className on only the parent element.

PR #1405, opened 2023-09-23
feat(generator): improved styled factory

astahmer:
## 📝 Description

Improved styled factory by adding a 3rd (optional) argument:

```ts
interface FactoryOptions<TProps extends Dict> {
  dataAttr?: boolean
  defaultProps?: TProps
  shouldForwardProp?(prop: string, isCssProperty: (prop: string) => boolean, variantKeys: string[]): boolean
}
```

- Setting `dataAttr` to true will add a `data-recipe="{recipeName}"` attribute to the element with the recipe name. This
  is useful for testing and debugging.

- `defaultProps` allows you to skip writing wrapper components just to set a few props. It also allows you to locally
  override the default variants or base styles of a recipe.

- `shouldForwardProp` allows you to customize which props are forwarded to the underlying element. By default, all props
  except recipe variants and style props are forwarded.

no


website/pages/docs/concepts/style-props.mdx
```diff
@@ -96,7 +96,59 @@ const Button = ({ children }) => (
...
+
+const App = () => {
+  return (
+    // In this case, you should keep the `size` naming
+    <CustomCircle circleSize="4" />
+  )
+}
+```
+
+The same principles apply to : all style props, recipe variants, and pattern props.
```
anubra266, line 130:
```suggestion
The same principles apply to all style props, recipe variants, and pattern props.

**Review by anubra266 — APPROVED** [review]

File: packages/generator/src/artifacts/solid-jsx/jsx.ts [hunk 2]
```diff
@@ -36,13 +63,62 @@ export function generateSolidJsxFactory(ctx: Context) {
...
+              return cx(css(cvaStyles), localProps.class, localProps.className)
+            }`
+          })
+          .run()}
 
         const classes = configOrCva.__recipe__ ? recipeClass : cvaClass
 
+        if (forwardedProps.className) {
+          delete forwardedProps.className
+        }
```
segunadebayo, line 116: Why are we deleting here?

packages/generator/src/artifacts/solid-jsx/jsx.ts
```diff
@@ -8,19 +9,45 @@ export function generateSolidJsxFactory(ctx: Context) {
...
+      const shouldForwardProp = (prop) => forwardFn(prop, cvaFn.variantKeys)
+      const initialProps = Object.assign(
+        options.dataAttr && configOrCva.__name__ ? { 'data-recipe': configOrCva.__name__ } : {},
+        options.defaultProps,
+      )
+
       return function ${componentName}(props) {
-        const mergedProps = mergeProps({ as: element }, props)
+        const mergedProps = mergeProps({ as: element }, initialProps, props)
+        const forwardedKeys = Object.keys(props).filter(shouldForwardProp)
```
segunadebayo, line 42: I think this needs to be wrapped in `createMemo` from solid-js so it's reactive

sandbox/vite-ts/src/App.tsx
```diff
@@ -1,13 +1,37 @@
...
 
+const StyledMotion = panda(
+  motion.div,
+  {},
+  {
+    /*  Allow motion props and non-style props to be forwarded. */
+    shouldForwardProp: (prop, variantKeys) =>
+      isValidMotionProp(prop) || (!variantKeys.includes(prop) && !isCssProperty(prop)),
+  },
+)
```
segunadebayo, line 20: For the types to work correctly, I think we should do the reverse. Create a motion component from a styled.div

```jsx
const StyledMotion = motion(styled.div)
```

This might make the example not useful tbh, but we can use a different example.

packages/generator/src/artifacts/solid-jsx/jsx.ts
astahmer: the `splitProps` fn coming from solid doesn't exactly work the same as ours, the `className` ended up in `localProps` AND `forwardedProps` instead of just `localProps`

so when it was later used in the `createComponent(Dynamic, mergeProps(forwardedProps` for some reason that className would have priority vs the `get class () { return classes() }` resulting in not all of the expected classes being attached to the element

Issue #1430, opened 2023-09-27
Unable to extend config utilities after 0.15.2 update


brendon1555:
### Description

My existing utilities config is throwing TS errors after updating to 0.15.2.
`Object literal may only specify known properties, and '$size' does not exist in type 'PropertyConfig'. ts(2322)`

Here is the utilities config
```ts
utilities: {
    extend: {
        $size: {
            values: "sizes",
            transform: (value: string) => ({
                width: value,
                height: value,
            }),
        },
    },
},
```

### Link to Reproduction

https://play.panda-css.com/OisuwgnIk5

### Steps to reproduce

1. Add a custom utility

### Panda CSS Version

0.15.2

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

This does appear to still function ok if you ignore the error, so seems to only be a typing issue.

astahmer:
my bad, sorry about that, yes it's just a type error

PR #1431, opened 2023-09-27
types: fix ExtendableUtilityConfig

astahmer:
Closes https://github.com/chakra-ui/panda/issues/1430

## 📝 Description

- Fix `ExtendableUtilityConfig` typings after a regression in 0.15.2 (due to
  https://github.com/chakra-ui/panda/pull/1410)
- Fix `ExtendableTheme` (specifically make the `RecipeConfig` Partial inside the `theme: { extend: { ... } }` object),
  same for slotRecipes

no

### DIFF
--- a/packages/types/src/utility.ts
+++ b/packages/types/src/utility.ts
@@ -55,6 +55,10 @@ export type UtilityConfig = {
   [property in LiteralUnion<CssProperty>]?: PropertyConfig
 }
 
-export type ExtendableUtilityConfig = UtilityConfig & {
-  extend?: UtilityConfig
+type UtilityConfigWithExtend = {
+  [pattern in LiteralUnion<CssProperty>]?: PropertyConfig | UtilityConfig | undefined
+}
+
+export type ExtendableUtilityConfig = UtilityConfigWithExtend & {
+  extend?: UtilityConfig | undefined
 }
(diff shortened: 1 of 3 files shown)


packages/types/src/utility.ts
```diff
@@ -55,6 +55,10 @@ export type UtilityConfig = {
...
 
-export type ExtendableUtilityConfig = UtilityConfig & {
-  extend?: UtilityConfig
+type UtilityConfigWithExtend = {
+  [pattern in LiteralUnion<CssProperty>]?: PropertyConfig | UtilityConfig | undefined
+}
+
+export type ExtendableUtilityConfig = UtilityConfigWithExtend & {
+  extend?: UtilityConfig | undefined
 }
```
segunadebayo, line 64: I'm confused about this type

astahmer: before:
Screenshot 2023-09-27 at 13 52 36

after:
Screenshot 2023-09-27 at 13 53 16

I'm guessing this is needed for TS to understand what values are allowed

this is actually closer to what's done in the other extendable types like `ExtendablePatterns`:
https://github.com/chakra-ui/panda/blob/fix/1430-extendable-utility/packages/types/src/config.ts#L69-L72

astahmer: the `extend` key will then constraint the mapped type

PR #1448, opened 2023-09-28
feat(cli): cssgen --minimal / cssgen of type

astahmer:
## --minimal flag

Adds a new `--minimal` flag for the CLI on the `panda cssgen` command to skip generating CSS for theme tokens,
preflightkeyframes, static and global css

Thich means that the generated CSS will only contain the CSS related to the styles found in the included files.

> Note that you can use a `glob` to override the `config.include` option like this:
> `panda cssgen "src/**/*.css" --minimal`

This is useful when you want to split your CSS into multiple files, for example if you want to split by pages.

Use it like this:

```bash
panda cssgen "src/**/pages/*.css" --minimal --outfile dist/pages.css
```

---

## cssgen {type}

In addition to the optional `glob` that you can already pass to override the config.include option, the `panda cssgen`
command now accepts a new `{type}` argument to generate only a specific type of CSS:

- preflight
- tokens
- static
- global
- keyframes

> Note that this only works when passing an `--outfile`.

You can use it like this:

```bash
panda cssgen "static" --outfile dist/static.css
```

no

### DIFF
--- a/.changeset/many-apes-love.md
+++ b/.changeset/many-apes-love.md
@@ -1,0 +1,44 @@
+---
+'@pandacss/generator': patch
+'@pandacss/node': patch
+'@pandacss/dev': patch
+---
+
+## --minimal flag
+
+Adds a new `--minimal` flag for the CLI on the `panda cssgen` command to skip generating CSS for theme tokens,
+preflightkeyframes, static and global css
+
+Thich means that the generated CSS will only contain the CSS related to the styles found in the included files.
+
+> Note that you can use a `glob` to override the `config.include` option like this:
+> `panda cssgen "src/**/*.css" --minimal`
+
+This is useful when you want to split your CSS into multiple files, for example if you want to split by pages.
+
+Use it like this:
+
+```bash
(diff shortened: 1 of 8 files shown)


anubra266:
Let's add the docs here https://panda-css.com/docs/references/cli#panda-cssgen @astahmer

.changeset/many-apes-love.md
```diff
@@ -0,0 +1,44 @@
+---
+'@pandacss/generator': patch
+'@pandacss/node': patch
+'@pandacss/dev': patch
```
segunadebayo, line 4:
```suggestion
'@pandacss/generator': minor
'@pandacss/node': minor
'@pandacss/dev': minor
```

subvertallchris:
Any chance there's a complete example that demonstrates using this to split by pages in a Next.js project? We're currently using the default build settings and Panda runs as a result of calling `npm ci`. This PR and its related docs are straightforward enough but there are gaps in my understanding of how to go from the defaults to what's demonstrated here.

PR #1469, opened 2023-10-02
build: improve DX with custom TS conditions

astahmer:
## Public changes

> These changes are only relevant if you are directly using **other** Panda `@pandacss/xxx` packages than the
> `@pandacss/dev`.

Use predefined interfaces instead of relying on automatic TS type inference or type aliases. This should result in
snappier

This should fix issues with the generation of typescript declaration (`.d.ts`) files when using `@pandacss/xxx` packages
directly, such as:

```
src/config.ts(21,14): error TS2742: The inferred type of 'tokens' cannot be named without a reference to '../node_modules/@pandacss/types/src/shared'. This is likely not portable. A type annotation is necessa…
```

## Internal changes:

Use [`tsconfig.compilerOptions.customConditions`](https://www.typescriptlang.org/tsconfig#customConditions) to get live
type inference from our workspace packages without the need for the `dev` command running.

This allows editing any source file and seeing the result instantly or using cmd+click on any type will jump to the
source instead of dist/xxx.d.ts !

Each workspace package imports are automatically resolved from the sources, type-wise in the editor and runtime-wise
with [vite](in dev only = in this repo cause we use the tsconfig customConditions with development)

This relies on node custom conditions in package.json `exports`:
https://nodejs.org/api/packages.html#conditional-exports


packages/generator/src/engines/base.ts
```diff
@@ -133,3 +133,29 @@ export const getBaseEngine = (conf: ConfigResultWithHooks) => {
...
+  isTemplateLiteralSyntax: boolean
+  studio: RequiredBy<NonNullable<ConfigResultWithHooks['config']['studio']>, 'outdir'>
+  hash: {
+    tokens: boolean | undefined
+    className: boolean | undefined
+  }
+  prefix: {
+    tokens: string | undefined
+    className: string | undefined
+  }
```
segunadebayo, line 147: Let's make the types package the source of truth for these types.

I'd imagine HashOptions, PrefixOptions type

packages/generator/src/engines/index.ts
```diff
@@ -1,48 +1,70 @@
...
-        return `export * from '${this.__extDts(file)}';`
-      },
-      isTypeFile(file: string) {
-        return file.endsWith('.d.ts') || file.endsWith('.d.mts')
-      },
+    file: getFileEngine(config),
+  }
+}
+
+const getFileEngine = (config: UserConfig): PandaFileEngine => {
```
segunadebayo, line 18: Let's move to this another file

packages/generator/src/generator.ts
```diff
@@ -81,4 +88,30 @@ export const createGenerator = (conf: ConfigResultWithHooks) => {
...
+  messages: ReturnType<typeof getMessages>
+  parserOptions: {
+    importMap: InternalImportMap
+    jsx: {
+      framework: Context['jsx']['framework']
+      factory: Context['jsx']['factoryName']
+      styleProps: Context['jsx']['styleProps']
+      isStyleProp: Context['isValidProperty']
+      nodes: Array<Context['patterns']['details'][number] | Context['recipes']['details'][number]>
+    }
```
segunadebayo, line 109: I imagine we can do something like JsxOptions

packages/generator/src/engines/jsx.ts
```diff
@@ -12,3 +12,12 @@ export const getJsxEngine = (config: UserConfig) => {
...
   }
 }
+
+export interface PandaJsxEngine {
+  factoryName: string
+  upperName: string
+  typeName: string
+  componentName: string
+  framework: UserConfig['jsxFramework']
+  styleProps: 'all' | 'minimal' | 'none'
```
segunadebayo, line 22: This should reference some type

PR #1492, opened 2023-10-05
feat: css generation mode grouped

astahmer:
https://panda-css.canny.io/feature-request/p/support-single-class-output-mode

this PR is just an exploration of how it could be done, if it is interesting and possible.

---

so far, this could be possible.. but mostly for simple cases, pretty much only when using `css()` with fully statically analyzable props.
some things just can't work with that `mode: "grouped`: `cva`, `sva`, merging styles with `.raw`, probably even JSX style props

it could work to some extent but whenever you'd be using some dynamic styling (some prop, value or condition that the static analysis can't find) things just wouldnt work, because classNames wouldn't naturally compose together

instead, some CSS rules would be generated with an incomplete list of prop/values/conditions when in the runtime a different className would be generated with something that the static analysis didnt catch

ex:

```
// mode: 'grouped'
const [color, setColor] = useState("red.400")
const className = css({ fontSize: "xl", px: 2, color })
// RUNTIME would produce something like: `fs_xl__px_2_color_red_400`
// STATIC ANALYSIS would produce something like: `fs_xl__px_2`
```


astahmer:
added cva here https://github.com/chakra-ui/panda/commit/52557404b99c8fa1372d72b42bef71a73c5acabc

Issue #503, opened 2023-04-25
[perf]: replacing postcss with lightningcss


astahmer:
the next big improvement that we can have performance-wise would be to replace [postcss](https://postcss.org/) with [lightningcss](https://lightningcss.dev/)

panda currently uses [those postcss plugins](https://github.com/chakra-ui/panda/blob/main/packages/core/package.json#L28-L34):

```json
"autoprefixer": "10.4.14",
"postcss": "8.4.23",
"postcss-discard-duplicates": "^6.0.0",
"postcss-discard-empty": "^6.0.0",
"postcss-merge-rules": "^6.0.0",
"postcss-nested": "6.0.1",
"postcss-normalize-whitespace": "^6.0.0",
"postcss-selector-parser": "^6.0.11",
```

using [`lightningcss`](https://lightningcss.dev/) we:

## CAN safely remove

- ✅ [autoprefixer](https://www.npmjs.com/package/autoprefixer)
> PostCSS plugin to parse CSS and add vendor prefixes to CSS rules using values from Can I Use. It is recommended by Google and used in Twitter and Alibaba.

we could replace it using [`postcss-lightningcss`](https://github.com/onigoetz/postcss-lightningcss), it's explicitly listed [in this list](https://github.com/onigoetz/postcss-lightningcss#postcss-plugins-that-you-can-remove-if-you-have-lightningcss) of postcss plugins that can be replaced

---

- ✅ [postcss-normalize-whitespace](https://www.npmjs.com/package/postcss-normalize-whitespace)
> Normalize whitespace with PostCSS.

[...]

PR #1544, opened 2023-10-19
refactor: internals, lightningcss and overall perf optimizations

astahmer:
## CSS generation
- rm chunks (outdir/chunks/xxx.css)
- sort rules before generating CSS (instead of previous approach of sorting after all CSS has been generated using a postcss plugin)
- insert rules in the ideal order by default, (maybe unlocking the possibility to remove @layer rules ?)
- hash each style objects into unique string to de-duplicate CSS rules and only generate them at most once (instead of previous approach of removing duplicates after all CSS has been generated using a postcss plugin)
- share the Stylesheet context between the different generator engines (global, reset, tokens, parser, static css etc)
- change `optimizeCss` implementation from postcss plugins to lightningcss

## Static CSS
- config.staticCss refactoring + StaticCss engine + static css for patterns and slot recipes
- new RuleProcessor to unify the style rules creation

## Fixes
- fix logger errors, print the stack trace as well instead of swallowing the error entirely
- fix `indeterminate` typo in `@pandacss/preset-base` conditions

## Misc
- refactor internals, move tests in appropriate packages, remove dependencies on `@pandacss/fixture` in each package and unify fixtures, only export the fixturePreset / fixture defaults config, remove some deadcode
- panda ship add a `schemaVersion` in the generated json

no


astahmer:
https://github.com/chakra-ui/panda/issues/503

packages/generator/__tests__/create-anatomy.ts
```diff
@@ -0,0 +1,54 @@
+// https://github.com/chakra-ui/zag/blob/215df8c66e953de02904c8419c1650f344e28faa/packages/anatomy/src/create-anatomy.ts
```
segunadebayo, line 1: Why not install zag-js anatomy for this one?

astahmer: cause it's just for tests and I didnt want to have the full selectors

https://github.com/chakra-ui/panda/pull/1544/files/ccf4825779a2a6855772a81370cadc496dcc4603..b5d8245cc937744aadcaa6c9a2ceb222ca52bd9b#diff-c836854eabc759fe5cdbff040b25cef0db437818faf43c9ad507b5294546432dR35-R39

segunadebayo:
Moving this to draft since it'll never be merged.

segunadebayo:
Closing this since we only use it as reference

PR #1546, opened 2023-10-19
opti: reduce outdir tokens.mjs size

astahmer:
## 📝 Description

Reduce `outdir` tokens.mjs file size

no

## 📝 Additional Information

on a large project:
unminified+ungzipped, that's about 35% less kb sent: 77.08kb  -> 46.87kb
unminified+gzipped, that's about 5% less kb sent: 8.48kb -> 8.05kb

note that this is also only relevant if a user is importing the js `token` function to do dynamic stuff at runtime, things like:
```ts
import { token } from '../styled-system/tokens'

token('sizes.2')
```

otherwise that file should be treeshaken by most bundlers, so 0kb sent most of the time, at most <10kb (since you should always use min+gzip) if you need runtime theme tokens.


segunadebayo:
Closing based on conversation

Issue #1549, opened 2023-10-19
OmittedHTMLProps should not affect minimal and none jsxStyleProps


ivanbanov:
### Description

**jsxStyleProps: 'all'**
```tsx
const Image = styled('img')

<Image htmlWidth='100' /> // 𝙭 - added htmlwidth as attr
<Image width='100' />     // ✔ - TS fail and no attr added
```

**jsxStyleProps: 'minimal' | 'none'**
```tsx
const Image = styled('img')

<Image htmlWidth='100' /> // 𝙭 - used htmlwidth as attr
<Image width='100' />     // 𝙭 - added width correctly but TS fails
```

### Link to Reproduction

https://stackblitz.com/edit/github-lhhzk8-jagbrd?file=src%2Fapp.tsx,panda.config.ts

### JS Framework

React

### Panda CSS Version

0.16.0

PR #1552, opened 2023-10-19
fix(#1549): OmittedHTMLProps + jsxStyleProps minimal/none

astahmer:
Closes https://github.com/chakra-ui/panda/issues/1549

## 📝 Description

Change `OmittedHTMLProps` to be empty when using `config.jsxStyleProps` as `minimal` or `none`

Fixes https://github.com/chakra-ui/panda/issues/1549

no


packages/generator/src/artifacts/types/generated.ts
```diff
@@ -28,8 +28,22 @@ export function getGeneratedTypes(ctx: Context) {
     system: rewriteImports(
       match(ctx.jsx.styleProps)
         .with('all', () => system.content)
-        .with('minimal', () => system.content.replace(jsxStyleProps, 'export type JsxStyleProps = WithCss'))
-        .with('none', () => system.content.replace(jsxStyleProps, 'export type JsxStyleProps = {}'))
+        .with('minimal', () =>
```
ivanbanov, line 31: this seems to solve only part of the problem, the omitted props still need to be applied when used with `html*` prefix

astahmer: good catch ! fixed and added a test there

https://github.com/chakra-ui/panda/pull/1552/commits/35f75ead73e6056a913de54843182805e17e346e#diff-2e5a346ed5ecd3ed6d409a1b4696b1b01b68351f077dbbf3efba8fa4a931669eR288-R303

sandbox/codegen/__tests__/frameworks/solid.styled-factory.test.tsx
```diff
@@ -163,6 +163,24 @@ describe('styled factory - cva', () => {
...
+      <styled.div htmlWidth={123} height="123">
+        Click me
+      </styled.div>,
+    )
+
+    expect(container.firstChild).toMatchInlineSnapshot(`
+      <div
+        class="h_123"
+        htmlwidth="123"
+        width="123"
```
ivanbanov, line 178: there is some side-effect here, only `width` is expected

sandbox/codegen/__tests__/frameworks/qwik.styled-factory.test.tsx
```diff
@@ -137,7 +139,20 @@ describe('styled factory - cva', async () => {
...
+
+  test('html props', async () => {
+    const { render, screen } = await createDOM()
+    await render(
+      <styled.div htmlWidth={123} height="123">
+        Click me
+      </styled.div>,
+    )
+    const container = screen.querySelector('div')!
+    expect(container.outerHTML).toMatchInlineSnapshot('"<div htmlwidth=\\"123\\" class=\\"h_123\\">Click me</div>"')
```
ivanbanov, line 155: seems like qwik didnt follow the change too 🤷‍♂️

sandbox/codegen/__tests__/frameworks/solid.styled-factory.test.tsx
astahmer: noticed that as well, note sure this can be fixed as the `mergeProps` is from solid, not from us, it doesn't split keys, it includes them in each matching function

maybe @segunadebayo knows ?

PR #1556, opened 2023-10-20
New type VariantStyledProps to extract variants from a styled component

ivanbanov:
## 📝 Description

New type added to `types/jsx` to easily extract variant types from a styled component

No

### DIFF
--- a/.changeset/tricky-berries-begin.md
+++ b/.changeset/tricky-berries-begin.md
@@ -1,0 +1,22 @@
+---
+'@pandacss/generator': minor
+---
+
+Added a new type to extract variants out of styled components
+
+```tsx
+import { VariantStyledProps } from '../styled-system/types/jsx'
+
+const Button = styled('button', {
+  base: { color: 'black' },
+  variants: {
+    state: {
+      error: { color: 'red' },
+      success: { color: 'green' },
+    },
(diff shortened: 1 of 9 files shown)


.changeset/tricky-berries-begin.md
```diff
@@ -0,0 +1,22 @@
...
+  base: { color: 'black' },
+  variants: {
+    state: {
+      error: { color: 'red' },
+      success: { color: 'green' },
+    },
+  },
+});
+
+type ButtonVariantProps = VariantStyledProps<typeof Button>;
```
astahmer, line 20: small nit, I'd personally prefer having a naming a bit more consistent with what we have currently (`RecipeVariantProps`), what do you think of `InferStyledVariantProps` ?

the rest of the PR looks good to me !

astahmer: I'm voting for either `StyledVariantProps<typeof Button>` or both 🙋🗳️

ivanbanov: So I used `VariantStyledProps` because the HTML type that uses `StyledProps` as the suffix `HTMLStyledProps`, sounded more consistent, I rename to the suggested one 👌

PR #1618, opened 2023-10-31
Fix studio semantic colors

anubra266:
Fix issue where error is thrown for semantic tokens with raw values

### DIFF
--- a/.changeset/fresh-monkeys-smash.md
+++ b/.changeset/fresh-monkeys-smash.md
@@ -1,0 +1,5 @@
+---
+'@pandacss/studio': patch
+---
+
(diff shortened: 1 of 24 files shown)


packages/studio/virtual-panda.ts
```diff
@@ -12,6 +12,7 @@ function vitePlugin(configPath: string): PluginOption {
   async function loadPandaConfig() {
     const ctx = await loadConfigAndCreateContext({ configPath })
     config = ctx.config
+    console.log('ctx.config', ctx.config)
```
astahmer, line 15: this was left out

astahmer (approved):

PR #1629, opened 2023-11-01
fix(playground): next 500 with client component + get .d.ts from local files

astahmer:
should be faster


anubra266 (approved):

playground/package.json
```diff
@@ -4,8 +4,9 @@
   "private": true,
   "scripts": {
     "css:codegen": "pnpm panda codegen --silent",
+    "types:codegen": "tsx scripts/bundle-types.mts",
     "prepare": "pnpm db:generate && pnpm css:codegen",
-    "build": "next build",
+    "build": "pnpm types:codegen && next build",
```
segunadebayo, line 9: Do we need to run this in the dev command as well?

astahmer: no, Ideally this is only needed once after installing
but when I added it in the `prepare` script, it was causing issues because of the packages build order

PR #1635, opened 2023-11-01
fix(cli): -i syntax option and prettify output

astahmer:
## 📝 Description

Fix CLI interactive mode `syntax` question values and prettify the generated `panda.config.ts` file

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -59,6 +59,7 @@
     "pluralize": "8.0.0",
     "postcss": "^8.4.31",
     "preferred-pm": "^3.0.3",
+    "prettier": "^2.8.8",
     "ts-morph": "19.0.0",
     "ts-pattern": "5.0.5",
     "tsconfck": "^2.1.2"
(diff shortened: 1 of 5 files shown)


packages/node/package.json
```diff
@@ -59,6 +59,7 @@
     "pluralize": "8.0.0",
     "postcss": "^8.4.31",
     "preferred-pm": "^3.0.3",
+    "prettier": "^2.8.8",
```
segunadebayo, line 62: Can we use the latest version of prettier?

packages/node/src/setup-config.ts
```diff
@@ -54,7 +55,7 @@ export default defineConfig({
 })
     `
 
-    await fsExtra.writeFile(join(cwd, file), content)
+    await fsExtra.writeFile(join(cwd, file), prettier.format(content))
```
segunadebayo, line 58: The latest version of prettier returns a promise here. So let's be sure to await that.

packages/node/package.json
astahmer: we could but I'd prefer updating it a later point, at the same time as the other prettier usages across the monorepo


Screenshot 2023-11-03 at 13 49 09

packages/node/src/setup-config.ts
astahmer: since this only applies to v3+, I don't think I need to change anything currently
Screenshot 2023-11-03 at 13 51 16

will do that when we update to v3 ! 👍

PR #1636, opened 2023-11-02
feat(cli): debug --only-config

astahmer:
## 📝 Description

Add a `--only-config` flag for the `panda debug` command, to skip writing app files and just output the resolved config.

### DIFF
--- a/.changeset/lucky-beds-sin.md
+++ b/.changeset/lucky-beds-sin.md
@@ -1,0 +1,5 @@
+---
+'@pandacss/node': patch
+---
+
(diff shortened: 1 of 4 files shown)


.changeset/lucky-beds-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@pandacss/node': patch
+---
+
+Add a `--only-config` flag for the `panda debug` command, to skip writing app files and just output the resolved config.
```
segunadebayo, line 5: Not a fan of this kind of flag. We probably want a `--filter` flag to filter the result to only the specified glob/paths. If we need to have this, then it'll be built on top of the filter flag.

Issue #1637, opened 2023-11-02
Issue with grid pattern


cschroeter:
### Description

The `minChildWidth` in the `grid` pattern does not respect token values.

### Link to Reproduction

https://play.panda-css.com/PXUbigKYRt

PR #1643, opened 2023-11-05
fix(#1637) grid minChildWidth value

astahmer:
Closes https://github.com/chakra-ui/panda/issues/1637

## 📝 Description

Fix an issue with the `grid` pattern from @pandacss/preset-base (included by default), setting a minChildWidth wasn't
interpreted as a token value

Before:

```tsx
<div className={grid({ minChildWidth: '80px', gap: 8 })} />
// ✅ grid-template-columns: repeat(auto-fit, minmax(80px, 1fr))

<div className={grid({ minChildWidth: '20', gap: 8 })} />
// ❌ grid-template-columns: repeat(auto-fit, minmax(20, 1fr))
//                                                  ^^^
```

After:

```tsx
<div className={grid({ minChildWidth: '80px', gap: 8 })} />
// ✅ grid-template-columns: repeat(auto-fit, minmax(80px, 1fr))

<div className={grid({ minChildWidth: '20', gap: 8 })} />
// ✅ grid-template-columns: repeat(auto-fit, minmax(var(--sizes-20, 20), 1fr))
//                                                  ^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/.changeset/lazy-rice-move.md
+++ b/.changeset/lazy-rice-move.md
@@ -1,0 +1,29 @@
+---
+'@pandacss/preset-base': patch
+'@pandacss/studio': patch
+---
+
+Fix an issue with the `grid` pattern from @pandacss/preset-base (included by default), setting a minChildWidth wasn't
+interpreted as a token value
+
+Before:
+
+```tsx
+<div className={grid({ minChildWidth: '80px', gap: 8 })} />
+// ✅ grid-template-columns: repeat(auto-fit, minmax(80px, 1fr))
+
+<div className={grid({ minChildWidth: '20', gap: 8 })} />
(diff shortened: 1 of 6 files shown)


fabianmarz:
@astahmer, @segunadebayo, is it possible this fix broke the `columns` property of the grid utility? When adding the grid is shown as rows instead of columns. Seems the resolving to the `sizes` token which would result in a `rem` value.
Link to example:
https://play.panda-css.com/2JDVpb15yA

image

segunadebayo:
Good catch @fabianmarz,

I've pushed an update to fix the regression.

anubra266:
@fabianmarz Fix is released in v0.18.2

PR #1659, opened 2023-11-09
Expand _required condition to include data and ARIA attributes

trevorblades:
## 📝 Description

The way the docs for [Conditional Styles](https://panda-css.com/docs/concepts/conditional-styles#state) are written make me believe that I should be able to react to the presence of a `data-required` prop using the `_required` condition, however the selector for `_required` doesn't currently account for that usage.

The `_required` condition only targets elements with a truthy `required` prop.

## 🚀 New behavior

`_required` targets elements with `required`, `data-required`, or `aria-required` props.

No


anubra266:
@trevorblades could you update this in the docs too. 
https://panda-css.com/docs/concepts/conditional-styles#reference

PR #1710, opened 2023-11-23
fix(#1686): migrate @typescript-eslint to v6 (#1686)

shinGangan:
## 📝 Description

- Migrate `@typescript-eslint` from v5 to v6:
  - ref: https://typescript-eslint.io/blog/announcing-typescript-eslint-v6/

- `typescript-eslint` v6 is being used as an ESLint rule.
- `typescript-eslint` is breaking changes from v5 to v6, which needs to be absorbed.

## 🚀 New behavior

- remove `.eslintignore`
  - This is because Flat config does not support loading ignore patterns from .eslintignore files.
  - ref: https://eslint.org/docs/latest/use/configure/migration-guide#ignoring-files
- fix `.eslintrc`
  - Enable `recommended` and `stylistic`, but only the same rules as the v5 recommended rules.
- ESLint and related libraries are now versioned in `package.json` in the workspace root.

No.

### typescript-eslint

- https://typescript-eslint.io/blog/announcing-typescript-eslint-v6/
- https://github.com/typescript-eslint/typescript-eslint/discussions/6014

### ESLint

- https://eslint.org/docs/latest/use/configure/migration-guide


shinGangan:
@anubra266, thanks for helping me. 
I was having trouble with workflow permissions.

All the CI's that are failing are pnpm install jobs. So I think updating pnpm-lock.json will solve the problem.

packages/types/src/prop-type.ts
```diff
@@ -5,6 +5,7 @@ import type { CssProperties } from './system-types'
  * Shadowed export (in CLI): DO NOT REMOVE
  * -----------------------------------------------------------------------------*/
 
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
```
JoshuaKGoldberg, line 8: 🤔 if this is set to `"off"` in the `.eslintrc`, there shouldn't be a need to disable it inline in files.

ESLint's [`--report-unused-disable-directives`](https://eslint.org/docs/latest/use/command-line-interface#--report-unused-disable-directives) can help here.

.eslintrc
```diff
@@ -20,6 +37,27 @@
         "argsIgnorePattern": "^_",
         "ignoreRestSiblings": true
       }
-    ]
+    ],
+    /**
+     * This rules is set up to use the same rules as "@typescript-eslint/recommended" in v5.
+     * @see https://typescript-eslint.io/blog/announcing-typescript-eslint-v6/
+     */
```
JoshuaKGoldberg, line 44: 👋 typescript-eslint maintainer here - I just noticed this PR has been sitting for a little while. Would you like any help with it?

.eslintrc
JoshuaKGoldberg, line 1: 👋 typescript-eslint maintainer here - I just noticed this PR has been sitting for a little while. Would you like any help with it @shinGangan / co.?

anubra266: Hi @JoshuaKGoldberg; Please feel free to point out any suggestions on this PR.

shinGangan: I was too busy with work to work on the PR.

@JoshuaKGoldberg , @anubra266 , thanks for following me. 🙏

PR #1756, opened 2023-12-03
feat: strictTokens escape hatch for arbitrary values

astahmer:
Closes https://github.com/chakra-ui/panda/discussions/1001

## 📝 Description

Add an escape-hatch for arbitrary values when using `config.strictTokens`, by prefixing the value with `[` and suffixing
with `]`, e.g. writing `[123px]` as a value will bypass the token validation.

```ts

css({
  // @ts-expect-error TS will throw when using from strictTokens: true
  color: '#fff',
  // @ts-expect-error TS will throw when using from strictTokens: true
  width: '100px',

  // ✅ but this is now allowed:
  bgColor: '[rgb(51 155 240)]',
  fontSize: '[12px]',
})
```

no

## 📝 Additional Information

We've listened and more enhancements will come for the `config.strictTokens` DX, thank you for providing great feedback !

### DIFF
--- a/packages/core/src/utility.ts
+++ b/packages/core/src/utility.ts
@@ -1,4 +1,12 @@
-import { compact, hypenateProperty, isFunction, isString, memo, withoutSpace } from '@pandacss/shared'
+import {
+  compact,
+  getArbitraryValue,
+  hypenateProperty,
+  isFunction,
+  isString,
+  memo,
+  withoutSpace,
+} from '@pandacss/shared'
 import type { TokenDictionary } from '@pandacss/token-dictionary'
 import type { AnyFunction, Dict, PropertyConfig, PropertyTransform, UtilityConfig } from '@pandacss/types'
 import type { TransformResult } from './types'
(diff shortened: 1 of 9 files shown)


packages/core/src/utility.ts
```diff
@@ -386,10 +394,11 @@ export class Utility {
       return { className: '', styles: {} }
     }
     const key = this.resolveShorthand(prop)
+
     return compact({
       layer: this.configs.get(key)?.layer,
       className: this.getOrCreateClassName(key, withoutSpace(value)),
-      styles: this.getOrCreateStyle(key, value),
+      styles: this.getOrCreateStyle(key, getArbitraryValue(value)),
```
segunadebayo, line 401: Should we only apply this when strictTokens is explicitly set? or always?

packages/generator/__tests__/generate-prop-types.test.ts
```diff
@@ -636,12 +636,15 @@ describe('generate property types', () => {
...
-        : ConditionalValue<string | number>"
+
+        type FilterString<T> = T extends \`\${infer _}\` ? T : never;
+        type WithArbitraryValue<T> = T | \`[\${string}]\`
+
+        export type PropertyValue<T extends string> = WithArbitraryValue<(T extends keyof PropertyTypes
+          ? ConditionalValue<FilterString<PropertyTypes[T]>>
+          : T extends keyof CssProperties
+          ? ConditionalValue<FilterString<CssProperties[T]>>
+          : ConditionalValue<string | number>)>"
```
segunadebayo, line 647: This type is getting a bit harder to reason about. Could we split it into multiple types and compose?

PR #1767, opened 2023-12-04
fix(patterns): with jsxStyleProps set to minimal/none

astahmer:
## 📝 Description

Fix issue where style props wouldn't be properly passed when using `config.jsxStyleProps` set to `minimal` or `none`
with JSX patterns (`Box`, `Stack`, `Flex`, etc.)

no

### DIFF
--- a/packages/generator/src/artifacts/preact-jsx/pattern.ts
+++ b/packages/generator/src/artifacts/preact-jsx/pattern.ts
@@ -4,13 +4,27 @@ import { match } from 'ts-pattern'
 import type { Context } from '../../engines'
 
 export function generatePreactJsxPattern(ctx: Context, filters?: ArtifactFilters) {
-  const { typeName, factoryName } = ctx.jsx
+  const { typeName, factoryName, styleProps: jsxStyleProps } = ctx.jsx
 
   const details = ctx.patterns.filterDetails(filters)
 
   return details.map((pattern) => {
     const { upperName, styleFnName, dashName, jsxName, props, blocklistType } = pattern
     const { description, jsxElement = 'div' } = pattern.config
+
+    const restProps = match(props.length)
+      .with(0, () => 'const restProps = props')
+      .otherwise(() => `const { ${props.join(', ')}, ...restProps } = props`)
+
+    const styleProps = match(props.length)
+      .with(0, () => `${styleFnName}()`)
+      .otherwise(() => `${styleFnName}({${props.join(', ')}})`)
+
+    const cssProps = match(jsxStyleProps)
+      .with('all', () => 'styleProps')
+      .with('minimal', () => '{ css: styleProps }')
+      .with('none', () => '{}')
+      .exhaustive()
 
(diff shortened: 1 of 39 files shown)


packages/generator/src/artifacts/preact-jsx/pattern.ts
```diff
@@ -4,14 +4,28 @@ import { match } from 'ts-pattern'
...
 
   const details = ctx.patterns.filterDetails(filters)
 
   return details.map((pattern) => {
     const { upperName, styleFnName, dashName, jsxName, props, blocklistType } = pattern
     const { description, jsxElement = 'div' } = pattern.config
 
+    const restProps = match(props.length)
+      .with(0, () => 'const restProps = props')
+      .otherwise(() => `const { ${props.join(', ')}, ...restProps } = props`)
```
astahmer, line 17: only keep this line & remove the trailing comma instead of using that "with(0")

packages/generator/src/artifacts/preact-jsx/pattern.ts
```diff
@@ -4,14 +4,28 @@ import { match } from 'ts-pattern'
...
+
+    const styleProps = match(props.length)
+      .with(0, () => `${styleFnName}()`)
+      .otherwise(() => `${styleFnName}({${props.join(', ')}})`)
```
astahmer, line 21: same

packages/generator/src/artifacts/preact-jsx/pattern.ts
```diff
@@ -4,14 +4,20 @@ import { match } from 'ts-pattern'
...
 
   const details = ctx.patterns.filterDetails(filters)
 
   return details.map((pattern) => {
     const { upperName, styleFnName, dashName, jsxName, props, blocklistType } = pattern
     const { description, jsxElement = 'div' } = pattern.config
 
+    const cssProps = match(jsxStyleProps)
+      .with('all', () => 'styleProps')
+      .with('minimal', () => '{ css: styleProps }')
```
segunadebayo, line 17:
```suggestion
      .with('minimal', () => '{ css: mergeCss(styleProps, props.css) }')
```

We might need to merge. the incoming CSS from props with the style props

astahmer: you mean in the solid one, not preact then ?

segunadebayo: I mean in all frameworks (of course we need to test this) to make sure the use can still override the styles using props.css later on

astahmer: ah yes, I misread as `mergeProps` instead of `mergeCss`

I tried to make the changes, but doing so cause cyclic import issues I think, leading to `Maximum call stack size exceeded` with nuxt
`import { mergeCss } from '../css/css.mjs';`

not sure how to fix it since mergeCss needs the same `css` context

PR #1814, opened 2023-12-12
Open props: Fix conditonal varaibles in border radii

anubra266:
Fix conditonal varaibles in border radii

### DIFF
--- a/packages/preset-open-props/__tests__/borders.test.ts
+++ b/packages/preset-open-props/__tests__/borders.test.ts
@@ -58,22 +58,22 @@ describe('Open Prop preset transforms', () => {
           "value": "49% 51% 48% 52% / 57% 44% 56% 43%",
         },
         "conditional-1": {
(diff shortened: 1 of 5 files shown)


packages/preset-open-props/__tests__/borders.test.ts
```diff
@@ -58,22 +58,22 @@ describe('Open Prop preset transforms', () => {
           "value": "49% 51% 48% 52% / 57% 44% 56% 43%",
         },
         "conditional-1": {
-          "value": "clamp(0px, calc(100vw - 100%) * 1e5, var(--radius-1))",
+          "value": "clamp(0px, calc(100vw - 100%) * 1e5, var(--radii-1))",
```
segunadebayo, line 61: As discussed, we should use the token reference syntax for this, and every other place.

```js
clamp(0px, calc(100vw - 100%) * 1e5, {radii.1})
```

PR #1835, opened 2023-12-15
docs: improve navigation layout

anubra266:
New docs layout


segunadebayo:
Marking this as a draft because we need to make some changes to be fully usable and accessible.

PR #1836, opened 2023-12-17
Add auto import to playground

anubra266:
Add auto import to playground

https://github.com/chakra-ui/panda/assets/30869823/e3819751-3aa8-4e0c-ae5d-c5876eaa5d3f

### DIFF
--- a/playground/src/lib/auto-import.ts
+++ b/playground/src/lib/auto-import.ts
@@ -1,22 +1,27 @@
 import { Monaco as MonacoType } from '@monaco-editor/react'
 import * as Monaco from 'monaco-editor'
-import { Artifact } from '@pandacss/types'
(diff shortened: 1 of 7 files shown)


playground/src/lib/auto-import.ts
```diff
@@ -0,0 +1,218 @@
...
+      label: 'cva',
+      path: 'css',
+      documentation: 'cva',
+    },
+    {
+      label: 'sva',
+      path: 'css',
+      documentation: 'sva',
+    },
+    { label: 'token', path: 'tokens', documentation: 'token' },
```
astahmer, line 73: can you add `isCssProperty, splitCssProps` from `/jsx` ?

astahmer (approved):

PR #1843, opened 2023-12-17
fix(diffs): avoid mutating config object

astahmer:
## 📝 Description

there was a nasty infinite loop caused by the diff-engine always detecting the same changes reproduction here: https://github.com/animareflection/ui/pull/181

```
{
  artifacts: Set(3) { 'recipes-index', 'slot-recipes.drawer', 'recipes' },
  hasConfigChanged: true,
  diffs: [
    { type: 'CREATE', path: ["theme", "slotRecipes", dDrawer", "jsx"], value: 'Drawer.Trigger' },
    { type: 'CREATE', path: ["theme", "slotRecipes", dDrawer", "jsx"], value: 'Drawer.Backdrop' },
    { type: 'CREATE', path: ["theme", "slotRecipes", dDrawer", "jsx"], value: 'Drawer.Positioner' },
    { type: 'CREATE', path: ["theme", "slotRecipes", dDrawer", "jsx"], value: 'Drawer.Content' },
    { type: 'CREATE', path: ["theme", "slotRecipes", dDrawer", "jsx"], value: 'Drawer.Title' },
    { type: 'CREATE', path: ["theme", "slotRecipes", dDrawer", "jsx"], value: 'Drawer.Description' },
    { type: 'CREATE', path: ["theme", "slotRecipes", dDrawer", "jsx"], value: 'Drawer.CloseTrigger' }
  ]
}
🐼 debug [builder] ⚙️ Config changed, reloading
🐼 debug [write:file] recipes/index.mjs
🐼 debug [write:file] recipes/index.d.ts
🐼 debug [write:file] recipes/drawer.mjs
🐼 debug [write:file] recipes/drawer.d.ts
🐼 info [hrtime] Extracted in (88.99ms)
```

the main issue is that:
1. our diff engine serialize the resolved config right after loading/bundling it (when it's finalized: after each presets has been merged)
2. we add these defaults jsx component names for each slot recipes (so that users have their recipes "just work" most of the time without specifying anything) directly on the `jsx` array coming from the config, essentially mutating it

-> this will create diffs and will be seen as if they were added by an actual config change

---

by mutating on another object/array reference (cloned with Array.from or Object.assign) than those coming from the resolved config, we should avoid this problem

### DIFF
--- a/.changeset/thick-moles-flow.md
+++ b/.changeset/thick-moles-flow.md
@@ -1,0 +1,7 @@
+---
+'@pandacss/generator': patch
+'@pandacss/core': patch
+'@pandacss/node': patch
+---
+
+Fix an issue with config change detection when using a custom `config.slotRecipes[xxx].jsx` array
--- a/packages/core/src/recipes.ts
+++ b/packages/core/src/recipes.ts
@@ -108,7 +108,7 @@ export class Recipes {
   private assignRecipe = (name: string, recipe: RecipeConfig | SlotRecipeConfig) => {
     const variantKeys = Object.keys(recipe.variants ?? {})
     const capitalized = capitalize(name)
-    const jsx = recipe.jsx ?? [capitalized]
+    const jsx = Array.from(recipe.jsx ?? [capitalized])
     if ('slots' in recipe) {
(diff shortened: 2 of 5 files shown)


coopbri (approved):
Thanks a lot for implementing this Alex -- I can confirm this works locally

PR #1848, opened 2023-12-18
feat: formatClassName, formatTokenName, and formatCssVar

ivanbanov:
## 📝 Description

Migrating a large codebase from Stitches to PandaCSS imposed some challenges due to the naming convention differences between the libraries. The main complication comes from the `$` prefix in Stitches, the workaround suggested by PandaCSS is simply adding a `$` sign in the token name, but it causes some problems.

- The CSS var is also generated with the dollar sign, and there is no way to dasherize it to avoid breaking the current CSS vars that are added by stitches e.g
    - PandaCSS `--colors-\$red.100`
    - Stitches `--colors-red-100`
- The final class name is not optimal with the inflict dollar sign. eg. `.text_\$red.100`
- The token names can not be customized making it harder to migrate from a codebase using purely `-` or `_` for example in the css vars. It would be nice to have an option to modify how the final token name is generated.
    - PandaCSS `red.100`
    - Stitches `$red-100`

## ⛳️ Current behavior

It's not possible to customize the token names, class names, and cssvars

## 🚀 New behavior

- Added a new config `formatTokenName`
  This config receives the token path and returns a string with the new token name

- Added a new config `formatClassName`
  This config receives the token as a string (from the formatTokenName) and returns a string that will be used as `<prefix>-<formatted classname>`. Hashed class names are not affected

- Added a new config `formatCssVar`
  This config receives the token path and returns a formatted name. Hashed class names are not affected

all together

```tsx
const dasherize = (token) => token
 .toString()
 .replace(/[^a-zA-Z0-9]+/g, '-')
 .replace(/^-+|-+$/g, '')

export default defineConfig({
  // Stitches preset
  separator: '-',
  formatTokenName: (path) => `$${path.join('-')}`,
  formatClassName: dasherize,
  formatCssVar: dasherize,
})
```

No


website/pages/docs/references/config.md
```diff
@@ -748,6 +748,32 @@ Only allow token values and prevent custom or raw CSS values. Will only affect p
 }
 ```
 
+### formatTokenName
```
anubra266, line 751: Let's put this section below `strictPropertyValues` please

website/pages/docs/migration/stitches.md
```diff
@@ -292,16 +292,19 @@ const styles = css({
...
-```diff
 export default defineConfig({
-  theme: {
-    colors: {
--      gray100: { value: 'hsl(206,22%,99%)' },
-+      $gray100: { value: 'hsl(206,22%,99%)' },
-    },
-  }
+  // Stitches preset
+  separator: '-',
```
anubra266, line 304: If we already have `.join('-')` in formatTokenName, is this still neccessary?
cc @astahmer

ivanbanov: the separator sits between the class prefix and the final class, the `.join('-')` is meant to join the token path

anubra266:
Found a bug @ivanbanov when I set prefix in the config, the separator is automatically set to `_` meanwhile the default separator in panda is `-`

ivanbanov:
Hey @anubra266 Im not sure Im following 🤔 the default separator from [Panda is actually **_**](https://github.com/chakra-ui/panda/blob/main/packages/types/src/config.ts#L236C15-L236C16)

While switching the configs I also got stuck in the state that you show in the image, but after rebuilding everything was properly built, Im wondering if it was a glitch in the autoreload

ivanbanov:
hey @segunadebayo

when I first added the `formatClassName` I also thought about that but I preferred to go for the trade-off "_if you format it, you are responsible for not breaking it_", and although the example you shared is a possible clash I wonder how much it will be an issue in the real world.

For sure, adding `tokenNamePrefix` is already a big win, but if I understand correctly it would replace the `formatTokenName` and `formatCssVar` options too (which are used to dasherize in my example), right?

Just dropping other ideas that I had to walk around it
- throw an error when a class clashes with another one
- instead of instead of receiving a function in the formatters, only allow some predefined options like `dash | camelCase | escape` and then use the `tokenNamePrefix` (but I wouldn't add the prefix to the final css and classname)
e.g.
```js
// config
{
  theme: {
    colors: {
      red: { 200: { value: '<value>' } }
    }
  },
  tokenNamePrefix: '$',
  tokenFormat: 'dash' 
}

css({ color: '$red-200' })

/**
 * classname: .text_red-200
 * cssvar:    --var(--colors-red-200)
 * token:     $red-200
 */
```

would it be a possible middle-term in the approaches?

segunadebayo:
Good points @ivanbanov.

The goal of the CSS function is to join the property and value with `_`; it doesn't (and probably shouldn't) have knowledge of the token and token name prefix to keep the runtime fast.

Not adding the prefix to the final class name is primarily an aesthetic concern; I don't assume it'll be a deal breaker for most users. I recommend using the `hash: true` to make the classNames terser.

The `tokenFormat` sounds interesting. What would the other options look like? I mean the 'camelCase' and `escape` options.

ivanbanov:
@segunadebayo Actually before I started implementing the cssVar formatter as a function I first tried the predefined approach, which worked as expected, you can see it [here](https://github.com/chakra-ui/panda/blob/6880116439ff5707ed582bd052951ec72ce03fe6/packages/shared/src/css-var.ts#L25). I even thought it would be possible to do the same in with the className and tokenName, but instead I went for the 100% customization.

Going in this direction I'd  say that we could have the following API

```js
theme: {
  tokens: {
    spacing: {
      xLarge: { value: '12px' }
    }
  },
  tokenNamePrefix: '$',
  tokenFormat: <'dash' | 'camelCase' | 'snakeCase' | 'escaped'>
}


/**
 * dash
 * classname: .m_x-large
 * cssvar:    --var(--spacing-x-large)
 * token:     $x-large
 */

/**
 * camelCase
 * classname: .m_XLarge
 * cssvar:    --var(--spacingXLarge)
 * token:     $xLarge
 */

/**
 * snakeCase
 * classname: .m_x_large
 * cssvar:    --var(--spacing_x_large)
 * token:     $x_large
 */

/**
 * escaped
 * classname: .m_$xLarge
 * cssvar:    --var(--spacing-\$xLarge)
 * token:     $xLarge
 */
```

It would still achieve my main goal, and give less flexibility to the final user, which can be good to avoid class name clashes, how does it sound to you?

segunadebayo:
Sounds good to me @ivanbanov 

The `className` might not ship with this. I don't see a clean way to solve it because requires the `css` function to have knowledge of properties that could use token. We shouldn't do that.

If we consider another token category, say spacing, there should be a semantic difference between "using a token" or not.

```jsx
css({ margin: "$40" })
// class: m_$40
// css: margin: var(--spacing-40)

css({ margin: "40" })
// class: m_40
// css: margin: 40px
```

Customizing the className will be problematic in this use case.

I'd recommend scoping this down to configuring the token name and format. Coming from Stitches, the class names are hashed there and don't see much value in customizing it.

ivanbanov:
@segunadebayo I see it, the `formatCssVar` and `formatTokenName` are only connected to the tokens while the `formatClassName` is more flexible and has to "understand" arbitrary values.

Im fine removing it, do you think that the other formatters are in good shape to make their way into the `v1`? Which direction do you want to take from here?

PR #1855, opened 2023-12-19
feat(sva): automatically generate CSS for sva with unresolved slots

astahmer:
## 📝 Description

Automatically extract/generate CSS for `sva` even if `slots` are not statically extractable, since it will only produce
atomic styles, we don't care much about slots for `sva` specifically

Currently the CSS won't be generated if the `slots` are missing which can be problematic when getting them from another
file, such as when using `Ark-UI` like `import { comboboxAnatomy } from '@ark-ui/anatomy'`

```ts
import { sva } from '../styled-system/css'
import { slots } from './slots'

const card = sva({
  slots, // ❌ did NOT work -> ✅ will now work as expected
  base: {
    root: {
      p: '6',
      m: '4',
      w: 'md',
      boxShadow: 'md',
      borderRadius: 'md',
      _dark: { bg: '#262626', color: 'white' },
    },
    content: {
      textStyle: 'lg',
    },
    title: {
      textStyle: 'xl',
      fontWeight: 'semibold',
      pb: '2',
    },
  },
})
```

no

### DIFF
--- a/packages/core/src/stylesheet.ts
+++ b/packages/core/src/stylesheet.ts
@@ -72,16 +72,7 @@ export class Stylesheet {
     recipe: Pick<SlotRecipeConfig, 'base' | 'variants' | 'compoundVariants'> & Partial<Pick<SlotRecipeConfig, 'slots'>>,
   ) => {
     if (!recipe.slots) {
-      const slots = new Set<string>()
-      Object.keys(recipe.base ?? {}).forEach((name) => {
-        slots.add(name)
-      })
-
-      Object.keys(recipe.variants ?? {}).forEach((name) => {
-        slots.add(name)
-      })
-
-      recipe.slots = Array.from(slots)
+      recipe.slots = Array.from(inferSlots(recipe as any))
     }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/stylesheet.ts
```diff
@@ -68,8 +68,24 @@ export class Stylesheet {
...
+  processAtomicSlotRecipe = (
+    recipe: Pick<SlotRecipeConfig, 'base' | 'variants' | 'compoundVariants'> & Partial<Pick<SlotRecipeConfig, 'slots'>>,
+  ) => {
+    if (!recipe.slots) {
+      const slots = new Set<string>()
+      Object.keys(recipe.base ?? {}).forEach((name) => {
+        slots.add(name)
+      })
+
+      Object.keys(recipe.variants ?? {}).forEach((name) => {
```
segunadebayo, line 80: This might be a bug since recipe.variants represents each variant key, not slot key.

we probably need to get the keys of each variant to properly compute the slots. I’d recommend refactoring this to a “inferSlots” function.

PR #1860, opened 2023-12-20
perf: skip duplicate rules with hashing

### LINKED ISSUE #1883: Vite HMR issue since 0.22.0
### Description

Since `v0.22.0`, there seems to be a bug with PandaCSS + Vite dev server. 
I am unsure if it's Vite, Panda, or a combination, but downgrading to `v0.21.0` resolves the issue. 

**Issue:** 
Whenever I change a file, a style or anything that triggers a Hot Module Reload, the initial few HMRs are fast but become increasingly slower with every update. 

After 10-20 HMR, the browser hangs, trying to load the `index.css` file that contains the Panda layer rules. 

I've tried anything I could think of, downgrading several packages, including Vite, disabling Mock Service Worker and numerous other things, but downgrading to Panda `v.0.21.0` resolves the issue. 

I cannot serve a concrete reproduction of the issue, but I hope you have enough context to possibly debug what happened after `v0.21.0` or find that this is entirely unrelated to PandaCSS. 

**Project relevant deps:**
* Vite 5.0.10
* Turborepo 1.11.2
* Typescript 5.3.3

### Steps to reproduce

N/A. 

### JS Framework

React (TS)

### Panda CSS Version

0.22.0

### Browser

Chrome latest

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

CleanShot 2023-12-27 at 21 46 02@2x

astahmer:
## 📝 Description

continuing the work previously done in https://github.com/chakra-ui/panda/pull/1544

Closes #1883

## CSS generation
- hash each style objects into unique string to de-duplicate CSS rules and only generate them at most once (instead of previous approach of removing duplicates after all CSS has been generated using a postcss plugin)

## Static CSS
- config.staticCss refactoring + StaticCss engine
- new RuleProcessor to unify the style rules creation

## Misc
- move core engines/context to @pandacss/core, remove some deadcode


packages/core/src/style-collector.ts
```diff
@@ -0,0 +1,282 @@
...
+
+    if (i === path.length - 1) {
+      Object.assign(current[key], value)
+    } else {
+      current = current[key]
+    }
+  }
+
+  return obj
+}
```
segunadebayo, line 282: Can we move these out of the file? Either to the shared package or to a different files

packages/core/src/stylesheet.ts
```diff
@@ -1,174 +1,123 @@
...
+export type LayerName =
+  | 'base'
+  | 'reset'
+  | 'recipes_slots_base'
+  | 'recipes_base'
+  | 'tokens'
+  | 'recipes'
+  | 'utilities'
+  | 'recipes_slots'
+  | 'compositions'
```
segunadebayo, line 24: Can we move these to the './types' file

packages/node/src/builder.ts
```diff
@@ -139,47 +139,40 @@ export class Builder {
...
-    return ctx.getCss()
-  }
-
   isValidRoot = (root: Root) => {
     const ctx = this.getContextOrThrow()
     let valid = false
 
     root.walkAtRules('layer', (rule) => {
-      if (ctx.layers.isValidParams(rule.params)) {
+      if (ctx.isValidLayerParams(rule.params)) {
```
segunadebayo, line 152: I think keeping the `isValidParams` in the layer class is still a good idea. Don't think we should move it to the root, except I'm missing some context.

adaboese:
Looks like all tests are passing. Is it possible to get this released?

segunadebayo:
@adaboese, I've got a couple of CI failures to fix.

Once we're all green, I'll merge it, and we'll cut a new release.

adaboese:
Looks like this was merged and released, but the Panda 0.24.1 still has issue described in https://github.com/chakra-ui/panda/discussions/1866

astahmer:
got it ! can you provide a minimal repro ? you can try running with `PANDA_DEBUG=* pnpm panda debug` to isolate the slow files

adaboese:
Created a new issue so it does not get lost https://github.com/chakra-ui/panda/issues/1966

PR #1876, opened 2023-12-25
Playground: Config refactor and error handling

anubra266:
Playground/show config errors

### DIFF
--- a/playground/src/hooks/useConfig.ts
+++ b/playground/src/hooks/useConfig.ts
@@ -1,0 +1,39 @@
+import { Config } from '@pandacss/types'
+import { useEffect, useState } from 'react'
+import * as pandaDefs from '@pandacss/dev'
+
+export const useConfig = (_config: string) => {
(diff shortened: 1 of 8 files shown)


playground/src/hooks/useConfig.ts
```diff
@@ -0,0 +1,39 @@
+import { Config } from '@pandacss/types'
+import { useEffect, useState } from 'react'
+import * as pandaDefs from '@pandacss/dev'
+
+export const useConfig = (_config: string) => {
```
astahmer, line 5: might be out of scope for this PR but what do you think of supporting TS in the config ? currently it silently won't work, I'm thinking we could use esbuild-wasm for that or sucrase or some lightweight way to transpile it before the eval

astahmer (approved):

playground/src/hooks/useConfig.ts
anubra266: Makes sense to me @astahmer. Can work on this in a different PR

Issue #1955, opened 2024-01-07
CSS injection stops reloading when css file any rule in addition to `@layer` rules


kristojorg:
### Description

I made an `index.css` file that had my layer rules and some other styles. I was importing stylesheets for a framework, but any rule will demonstrate the problem:

```
@layer reset, base, tokens, recipes, utilities;

* {
  color: red;
}
```

When I include the extra css rule, the styles injected into the `<style>` tag seem to get stuck. That is to say, if you change the above to have `color: blue;`, it will not update even if you refresh the page, restart the server or run `panda codegen`. 

Interestingly, if I import my styles in Vite with `import styles from "./index.css?inline"` the `styles` import does in fact change with every change to the `index.css` file.

### Link to Reproduction

https://stackblitz.com/~/github.com/kristojorg/vite-bun-reload-problem

### Steps to reproduce

1. Go to the stackblitz
2. Run the app with the background color set in `panda.css`
3. Change the background color
4. The app will not reflect the new background color, even with a reload
5. Notice that if you remove the `@layer` rules, the styles will start updating normally when changed

### JS Framework

React TS

### Panda CSS Version

^0.25.0

### Browser

Arc 1.21.3 (but others also)

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

This started sometime between 0.18 and 0.25. I will try to start downgrading to see when it was introduced.

_No response_

PR #1956, opened 2024-01-07
fix(postcss): regression with user-defined rules not reloaded and kill astro plugin

astahmer:
Closes https://github.com/chakra-ui/panda/issues/1955

## 📝 Description

BREAKING: Remove the `@pandacss/astro` plugin in favor of just using the `postcss` plugin

Fix `@pandacss/postcss` plugin regression when the entry CSS file (with `@layer` rules order) contains user-defined
rules, those user-defined rules would not be reloaded correctly after being changed. (This was due to a `@pandacss/astro` specific fix that affected users of the postcss plugin)

---

Fix an edge-case for when the `config.outdir` would not be set in the `panda.config`

Internal details: The `outdir` would not have any value after a config change due to the fallback being set in the
initial config resolving code path but not in context reloading code path, moving it inside the config loading function
fixes this issue.

YES if you're using the `@pandacss/astro` plugin please migrate to just using the `@pandacss/postcss` plugin (you don't have to install anything, it's included with in `@pandacss/dev/postcss`) by taking [these steps](https://panda-css.com/docs/installation/vite#configure-the-entry-css-with-layers):

1. Remove the `@pandacss/astro` plugin usage from your `astro.config`
2. Configure the entry CSS with layers: create a `.css` file (or to an existing one) the `@layer` rules order (which was implicitly added by the `@pandacss/astro` before): `@layer reset, base, tokens, recipes, utilities;`, 
3. make sure to import this `.css` file (new or existing) somewhere in your app, it makes sense to add it in a `layout.astro` file

## 📝 Additional Information

the `@pandacss/astro` had little value, it was only a wrapper of the `@pandacss/postcss` plugin and only allowed to skip 1 step: importing a `.css` file with the `@layer` rules.
it created unnecessary maintenance (like the issue linked above) and specific tweaks were needed to make it work seamlessly that also affected other (non-astro) `postcss` users


kristojorg:
Nice, thank you!

kristojorg:
Can you tell me if this has been released? I am still seeing this bug. I specifically need to add a layer to my layer order and am unable to get panda to work when doing this:
```
@layer reset, ionic, base, tokens, recipes, utilities;
```

anubra266:
@kristojorg what method are you using to setup panda, and it would be helpful if you could attach a reproduction

kristojorg:
@anubra266 I have failed to reproduce on my reproduction stackblitz from my previous ticket with the newest panda package. I will debug my local setup and if I can reproduce, I'll open a new issue

Omikorin:
@kristojorg did you add postcss config from [these docs](https://panda-css.com/docs/installation/astro#update-the-postcss-config)? For me, it was a must-have to get CSS working again.

PR #1960, opened 2024-01-07
refactor: extract import logic

segunadebayo:
## 📝 Description

This PR refactors the parser and import analysis logic.
- The goal is to ensure we can re-use the import logic in ESLint Plugin and Node.js.
- It also allows us to isolate and cover the import analysis logic with many tests.
- Converts some ts-pattern `match` to `switch-case`

Hard to re-use the import logic in the parser

## 🚀 New behavior

Now isolated to the core package

No

## 📝 Additional Information

Closes #1962


packages/core/src/jsx.ts
```diff
@@ -0,0 +1,121 @@
...
+        } else {
+          this.patternMatcher.regex.push(jsx)
+        }
+      })
+    }
+  }
+
+  get names() {
+    return [this.factoryName, ...this.nodes.map((node) => node.jsxName)]
+  }
```
astahmer, line 58: this will get re-executed on each property access, we probably want to save it to a property class instead since it loops on each nodes

astahmer (approved):

PR #1969, opened 2024-01-08
refactor(hooks): rm hookable, use raw object + callbacks

astahmer:
## 📝 Description

Refactor `config.hooks` to be much more powerful, you can now:

- Tweak the config after it has been resolved (after presets are loaded and merged), this could be used to dynamically
  load all `recipes` from a folder
- Transform a source file's content before parsing it, this could be used to transform the file content to a
  `tsx`-friendly syntax so that Panda's parser can parse it.
- Implement your own parser logic and add the extracted results to the classic Panda pipeline, this could be used to
  parse style usage from any template language
- Tweak the CSS content for any `@layer` or even right before it's written to disk (if using the CLI) or injected
  through the postcss plugin, allowing all kinds of customizations like removing the unused CSS variables, etc.
- React to any config change or after the codegen step (your outdir, the `styled-system` folder) have been generated

See the list of available `config.hooks` here:

```ts
export interface PandaHooks {
  /**
   * Called when the config is resolved, after all the presets are loaded and merged.
   * This is the first hook called, you can use it to tweak the config before the context is created.
   */
  'config:resolved': (args: { conf: LoadConfigResult }) => MaybeAsyncReturn
  /**
   * Called when the Panda context has been created and the API is ready to be used.
   */
  'context:created': (args: { ctx: ApiInterface; logger: LoggerInterface }) => void
  /**
   * Called when the config file or one of its dependencies (imports) has changed.
   */
  'config:change': (args: { config: UserConfig }) => MaybeAsyncReturn
  /**
   * Called after reading the file content but before parsing it.
   * You can use this hook to transform the file content to a tsx-friendly syntax so that Panda's parser can parse it.
   * You can also use this hook to parse the file's content on your side using a custom parser, in this case you don't have to return anything.
   */
  'parser:before': (args: { filePath: string; content: string }) => string | void
  /**
   * Called after the file styles are extracted and processed into the resulting ParserResult object.
   * You can also use this hook to add your own extraction results from your custom parser to the ParserResult object.
   */
  'parser:after': (args: { filePath: string; result: ParserResultInterface | undefined }) => void
  /**
   * Called after the codegen is completed
   */
  'codegen:done': () => MaybeAsyncReturn
  /**
   * Called right before adding the design-system CSS (global, static, preflight, tokens, keyframes) to the final CSS
   * Called right before writing/injecting the final CSS (styles.css) that contains the design-system CSS and the parser CSS
   * You can use it to tweak the CSS content before it's written to disk or injected through the postcss plugin.
   */
  'cssgen:done': (args: {
    artifact: 'global' | 'static' | 'reset' | 'tokens' | 'keyframes' | 'styles.css'
    content: string
  }) => string | void
}
```

yes if you're currently using config.hooks, which aren't currently documented, so very unlikely


packages/core/src/hooks-api.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  get config() {
+    return this.ctx.conf.config
+  }
+
+  get configPath() {
+    return this.ctx.conf.path
+  }
+
+  get dependencies() {
```
segunadebayo, line 19:
```suggestion
  get configDependencies() {
```

packages/core/src/stylesheet.ts
```diff
@@ -44,7 +44,13 @@ export class Stylesheet {
 
   processGlobalCss = (styles: Dict) => {
     const result = this.serialize(styles)
-    this.context.layers.base.append(stringify(result))
+    let css = stringify(result)
+
+    if (this.context.hooks['cssgen:done']) {
+      css = this.context.hooks['cssgen:done']?.({ artifact: 'global', content: css }) ?? css
```
segunadebayo, line 50:
```suggestion
      css = this.context.hooks['cssgen:done']({ artifact: 'global', content: css }) ?? css
```

packages/generator/src/artifacts/css/keyframe-css.ts
```diff
@@ -10,8 +10,11 @@ export function generateKeyframeCss(ctx: Context, sheet: Stylesheet) {
     result[`@keyframes ${name}`] = definition
   }
 
-  const css = stringify(sheet.serialize(result))
+  let css = stringify(sheet.serialize(result))
+
+  if (ctx.hooks['cssgen:done']) {
+    css = ctx.hooks['cssgen:done']?.({ artifact: 'keyframes', content: css }) ?? css
```
segunadebayo, line 16:
```suggestion
    css = ctx.hooks['cssgen:done']({ artifact: 'keyframes', content: css }) ?? css
```

packages/generator/src/artifacts/css/static-css.ts
```diff
@@ -7,8 +7,12 @@ export const generateStaticCss = (ctx: Context, sheet?: Stylesheet) => {
 
   if (!sheet) {
     const { optimize = true, minify } = config
-    const output = engine.sheet.toCss({ optimize, minify })
-    void ctx.hooks.callHook('generator:css', 'static.css', output)
+    let output = engine.sheet.toCss({ optimize, minify })
+
+    if (ctx.hooks['cssgen:done']) {
+      output = ctx.hooks['cssgen:done']?.({ artifact: 'static', content: output }) ?? output
```
segunadebayo, line 13:
```suggestion
      output = ctx.hooks['cssgen:done']({ artifact: 'static', content: output }) ?? output
```

We could also rename `output` to `css`

PR #1983, opened 2024-01-09
docs: patterns JSX tab + built-in conditions link + rm @pandacss/astro from get started

astahmer:


website/pages/docs/customization/conditions.mdx
```diff
@@ -66,3 +66,46 @@ function App() {
...
+You can extend the [default conditions](/docs/concepts/conditional-styles#reference) by using the `conditions.extend` property in the config.
+
+
+```ts filename="panda.config.ts"
+import { defineConfig } from '@pandacss/dev'
+
+export default defineConfig({
+  conditions: {
+    extend: {
+      // Extend the default `flex` pattern
```
anubra266, line 81:
```suggestion
      // Extend the default `dark` condition

File: website/pages/docs/customization/conditions.mdx [hunk 2]
```diff
@@ -66,3 +66,46 @@ function App() {
...
+      dark: '.dark &, [data-theme="dark"] &',
+    },
+  },
+});
+```
+
+Then you can run the following command to update the pattern JS code:
```
anubra266, line 88:
```suggestion
Then you can run the following command to update the conditions JS code:

File: website/pages/docs/concepts/patterns.mdx [hunk 3]
```diff
@@ -12,6 +12,22 @@ predefined patterns like stack, hstack, vstack, wrap, etc. These patterns can be
 
 ## Predefined Patterns
 
+### Box
```
anubra266, line 15: There's no function/JSX tabs for `Box`

anubra266: Oh...that makes sense

website/pages/docs/concepts/patterns.mdx
```diff
@@ -329,14 +641,31 @@ The `square` function accepts the following properties:
...
+
+function App() {
+  return <Square size="12" bg="red.300" />
+}
+```
+
+  </Tab>
+</Tabs>
+
 ### Visually Hidden
```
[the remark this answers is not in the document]
astahmer: I couldnt think of an example where it makes sense to use it as JSX, it would require using the `as` prop which is not strongly-typed

anubra266 (approved):

PR #1986, opened 2024-01-10
perf: improve runtime performance

astahmer:
Closes https://github.com/chakra-ui/panda/discussions/1982

## 📝 Description

Improve the performance of the runtime transform functions by caching their results (css, cva, sva, recipe/slot recipe,
patterns)

since Panda has to transform style objects to classes (`{ backgroundColor: "red.300" }` -> `bg_red`, it will always have a little bit of overhead vs just writing the final classes, caching those reduce the time spent on it.

no

## 📝 Additional Information

if you're using anything that can pre-render to HTML (Astro, RSC, etc) you can avoid this issue entirely by automatically inlining classes at build-time, without ever needing those runtime functions


segunadebayo:
Great work! Did some performance comparisons, rendering 100 instances of a component using cva.

**Before**
CleanShot 2024-01-11 at 15 45 21@2x

**After**
CleanShot 2024-01-11 at 15 45 39@2x


**Diff**
Mount Time: 10.9s -> 3.6s (100.69% improvement)
Update Time: 9s -> 4.8s (60.8696% difference)

PR #2016, opened 2024-01-16
Feature: ESLint plugin 🐼

anubra266:
Enforce consistent usage of tokens and CSS patterns across your codebase.


website/pages/docs/references/eslint.mdx
JoshuaKGoldberg, line 1: [Suggestion] https://github.com/bmish/eslint-doc-generator is pretty great and can help set up docs. I haven't looked into integrating it with an mdx site but would be surprised if it either is already supported well or something they'd be open to receiving an issue for.

 Totally not suggesting as a blocker for this PR.

website/pages/docs/references/eslint.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+Prefer atomic properties over shorthand properties.
+
+## Settings
+
+### `configPath`
+
+You can tell `eslint` to use a custom panda config file by setting the `configPath` option in your `.eslintrc.js` file.
+
+By default we find the nearest oanda config to the linted file.
```
JoshuaKGoldberg, line 63:
```suggestion
By default we find the nearest panda config to the linted file.
```

website/pages/docs/references/eslint.mdx
```diff
@@ -0,0 +1,74 @@
...
+    ```
+  </Tab>
+  <Tab>
+    ```bash
+    bun add -D @pandacss/eslint-plugin
+    ```
+  </Tab>
+</Tabs>
+
+## Rules
```
JoshuaKGoldberg, line 35: [Suggestion] As someone who is very new to Panda, I don't know what most of these mean by just the one sentence. It'd be nice to have more explanation from them. I'm guessing that's better as a follow-up task. 😄

anubra266: Yeah...I just put something there for now. We definitely have to flesh them up.

website/pages/docs/references/eslint.mdx
```diff
@@ -0,0 +1,74 @@
...
+    yarn add -D @pandacss/eslint-plugin
+    ```
+  </Tab>
+  <Tab>
+    ```bash
+    bun add -D @pandacss/eslint-plugin
+    ```
+  </Tab>
+</Tabs>
+
```
JoshuaKGoldberg, line 34: [Suggestion] IME it's good to also mention how to include the plugin in the ESLint config. Lots of folks are so low confidence with ESLint they'll be confused why the package manager installations on their own aren't working.

```suggestion

Something something then in your ESLint config:

\`\`\`json
"plugins": ["@pandacss/eslint-plugin"]
\`\`\`

```

If folks do see just the code block below, there's a good chance they'll skim, miss the actual explanation, and think `settings` are necessary.

website/pages/docs/references/eslint.mdx
```diff
@@ -0,0 +1,74 @@
...
+```json
+{
+  "plugins": [
+    "@pandacss/eslint-plugin"
+  ],
+  "settings": {
+    "pandacss/configPath": "/ABSOLUTE-PATH-TO/panda.config.js"
+  }
+}
```
JoshuaKGoldberg, line 83: [Bug] Hardcoded paths in ESLint aren't a common practice because they're often not predictable. E.g. CI machines will have different absolute paths than local configs.

The occasional need to do something like `path.join` is why a lot of folks tend to write ESLint configs as `.js`:

```suggestion
```js
const path = require("path");

module.exports = {
  plugins: [
    "@pandacss/eslint-plugin"
  ],
  settings: {
    "pandacss/configPath": path.join("PATH-TO/panda.config.js")
  }
};
```

website/pages/docs/references/eslint.mdx
```diff
@@ -0,0 +1,74 @@
...
+```
```
JoshuaKGoldberg, line 84: [Suggestion] Have you considered also adding a tab for the [new flat config format](https://eslint.org/docs/latest/use/configure/configuration-files-new)? We don't yet officially support it in typescript-eslint (long story) but a lot of consumers -including folks who newly set up their configs starting next month or so onward- will be defaulting to it.

website/pages/docs/references/eslint.mdx
anubra266: @JoshuaKGoldberg This would be a `eslint.config.js` config right?

website/pages/docs/references/eslint.mdx
anubra266: @JoshuaKGoldberg I thought of it, but I'm not sure it's setup to work with flat configs yet. Also yet to add `configs` i.e. recommended.

website/pages/docs/references/eslint.mdx
JoshuaKGoldberg: No, this code block is for the current/legacy-style `.eslintrc.(c)js`. That's what I personally use.

JoshuaKGoldberg:
I finished commenting btw - this is looking great! Thanks for setting it up! 🙌

anubra266:
> I finished commenting btw - this is looking great! Thanks for setting it up! 🙌

Thanks a lot for your review @JoshuaKGoldberg

.changeset/small-dogs-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@pandacss/eslint-plugin': minor
+---
+
+Add Panda Eslint Plugin
```
astahmer, line 5: this is probably WIP but this could use some instructions on how to install etc

packages/eslint-plugin/src/rules/file-not-included.test.ts
```diff
@@ -0,0 +1,31 @@
+import { tester } from 'src/test-utils'
```
astahmer, line 1: is this import intended ?

packages/eslint-plugin/src/rules/no-unsafe-token-fn-usage.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const invalids = [
+  'const styles = css({ bg: token("colors.red.300") })',
+  `const styles = css({ bg: 'token("colors.red.300")' })`,
+  '<div className={css({ border: "{borders.1}" })} />',
+  '<Circle _hover={{ color: "{colors.blue.300}" }} />',
+  '<Circle bg={token("colors.red.300")} />',
+  `<Circle bg={'token("colors.red.300")'} />`,
+  `<Circle bg='token("colors.red.300")' />`,
+]
```
astahmer, line 21: thinking about a different rule but a rule that checks for the validity of token paths (to prevent typos) is probably useful (we have something like that in the vscode ext)

packages/eslint-plugin/src/utils/helpers.ts
```diff
@@ -0,0 +1,170 @@
...
+  return node.type === 'JSXIdentifier' && isValidProperty(node.name, context)
+}
+
+const isPandaIsh = (name: string, context: RuleContext<any, any>) => {
+  const imports = getImports(context)
+  return syncAction('matchFile', getSyncOpts(context), name, imports)
+}
+
+// TODO - We might want to parse the AST from the root to find the declaration
+const findDeclaration = (name: string, context: RuleContext<any, any>) => {
```
astahmer, line 70: I'm not sure if those comments are still relevant but I implemented something related in the extractor 
https://github.com/chakra-ui/panda/blob/main/packages/extractor/src/find-identifier-value-declaration.ts#L78

packages/eslint-plugin/src/rules/no-hardcoded-color.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  '<Circle _hover={{ bgColor: "gray.100" }} />',
+]
+
+const invalids = [
+  'const styles = css({ color: "skyblue" })',
+  '<div className={css({ background: "#111" })} />',
+  '<Circle _hover={{ borderColor: "rgb(1, 1, 1)" }} />',
+  '<Circle _hover={{ bg: "hsl(0deg, 0%, 7%)" }} />',
+  '<Circle _hover={{ bgColor: "rgba(17, 17, 17, 1)" }} />',
+]
```
astahmer, line 23: we should allow using the `[xxx]` escape-hatch syntax

packages/eslint-plugin/src/rules/no-shorthand-prop.ts
```diff
@@ -0,0 +1,56 @@
+import { isPandaAttribute, isPandaProp, resolveLonghand } from '../utils/helpers'
+import { type Rule, createRule } from '../utils'
+
+export const RULE_NAME = 'no-shorthand-prop'
```
astahmer, line 4: any reason not to directly set config.shorthands to false ?

packages/eslint-plugin/src/rules/file-not-included.test.ts
anubra266: @astahmer Yeah...this is what we use in testing. Or did you mean something else?

packages/eslint-plugin/src/rules/no-hardcoded-color.test.ts
anubra266: Might need to discuss this more. But IMO, that's exactly something we want eslint to prevent us from doing. 
cc @segunadebayo

packages/eslint-plugin/src/rules/no-shorthand-prop.ts
anubra266: Hmm...this actually makes sense.
cc @segunadebayo

packages/eslint-plugin/src/rules/file-not-included.test.ts
anubra266: Oh...it didn't satisfy typescript

packages/eslint-plugin/src/rules/no-shorthand-prop.ts
anubra266: Agreed to keep this for Incremental adoption

anubra266:
Closing in favour of separate repo 
https://github.com/chakra-ui/eslint-plugin-panda

PR #2021, opened 2024-01-17
chore: tsup script cleanup

anubra266:
Cleanup tsup builds with `tsup config` file


packages/config/tsup.config.ts
```diff
@@ -1,7 +1,7 @@
 import { defineConfig } from 'tsup'
 
 export default defineConfig({
-  entry: ['src/index.ts', 'src/merge-config.ts', 'src/diff-config.ts', 'src/resolve-ts-path-pattern.ts'],
+  entry: ['src/index.ts', 'src/merge-config.ts', 'src/resolve-ts-path-pattern.ts'],
```
segunadebayo, line 4: We need to keep `src/diff-config.ts`. Please revert

PR #2032, opened 2024-01-19
💄 Revert em style and strong

k-utsumi:
## 📝 Description

Changed because italics were not applied to the markdown generation page (Storybook).
I set `revert` because I think setting `revert` is better for accessibility than setting any other value.

Italics were not applied.

## 🚀 New behavior

Italics will applied.

Which one...? 🤔 

## 📝 Additional Information

**Can I use...**

- CSS revert value  
  https://caniuse.com/css-revert-value
- `@layer` (used in Panda CSS)
  https://caniuse.com/css-cascade-layers


segunadebayo:
This is slightly opinionated. I’d recommend you add to your app instead.

PR #2048, opened 2024-01-20
feat: add support for token references with curly braces in media queries

astahmer:
## 📝 Description

Add support for token references with curly braces like `{path.to.token}` in media queries, just like the
`token(path.to.token)` alternative already could.

```ts
css({
  // ✅ this is fine now, will resolve to something like
  // `@container (min-width: 56em)`
  '@container (min-width: {sizes.4xl})': {
    color: 'green',
  },
})
```

Fix an issue where the curly token references would not be escaped if the token path was not found.

no


packages/core/src/style-decoder.ts
```diff
@@ -175,7 +176,9 @@ export class StyleDecoder {
     const sorted = sortStyleRules(details)
     sorted.forEach((value) => {
       if (value.conditions) {
-        const path = basePath.concat(value.conditions.map((c) => c.rawValue ?? c.raw))
+        const path = basePath.concat(
+          value.conditions.map((c) => this.context.utility.tokens.resolveReference(c.rawValue ?? c.raw)),
+        )
```
segunadebayo, line 181: We could create a `resolveCondition` method in the style decoder

this.context.utility.tokens.resolveReference(c.rawValue ?? c.raw)),`

packages/core/src/style-decoder.ts
```diff
@@ -112,7 +112,9 @@ export class StyleDecoder {
 
     if (entry.cond) {
       conditions = this.context.conditions.sort(parts)
-      const path = basePath.concat(conditions.map((c) => c.rawValue ?? c.raw))
+      const path = basePath.concat(
+        conditions.map((c) => this.context.utility.tokens.resolveReference(c.rawValue ?? c.raw)),
+      )
```
segunadebayo, line 117: Update this to use the new method

PR #2056, opened 2024-01-21
refactor: preserve underlines while hashing and decrease hashing length

theMasix:
## 📝 Description

We wanted to keep the last className of the same CSS property passed to a component and remove the other ones. So we decided to use a different method than `cx` that does this for us, we developed a method like [`tailwind-merge`](https://www.npmjs.com/package/tailwind-merge) but for PandaCSS.

The algorithm was based on splitting className by `_` and then keeping the last className of a CSS property and removing the others.
So the bottleneck was that the algorithm was useless while we activated `hash` on production because all `_` became hashed as well!

The hashing had been done regardless of the `_` between the className.

## 🚀 New behavior

the className is split based on `_` and then hash each part of it separately.
I also decreased the length of the hash strings since we now have two hash strings per class that are joined together at the end.

No. hashing is an internal process and the end-user should not change anything manually.

### DIFF
--- a/packages/core/__tests__/prefix.test.ts
+++ b/packages/core/__tests__/prefix.test.ts
@@ -14,22 +14,22 @@ describe('atomic-rule / prefix', () => {
   test('should product consistent hash', () => {
     expect(backend({ color: 'red' })).toMatchInlineSnapshot(`
       "@layer utilities {
-        .tw-eIKWVi {
+        .tw-bgaoI_ftuE {
           color: red;
       }
       }"
     `)
-    expect(frontend({ color: 'red' })).toMatchInlineSnapshot('"tw-eIKWVi"')
+    expect(frontend({ color: 'red' })).toMatchInlineSnapshot('"tw-bgaoI_ftuE"')
 
     expect(backend({ color: { sm: 'red' } })).toMatchInlineSnapshot(`
       "@layer utilities {
         @media screen and (min-width: 40em) {
-          .tw-geqOyW {
+          .tw-KGNB_ftuE {
             color: red;
       }
       }
       }"
     `)
-    expect(frontend({ color: { sm: 'red' } })).toMatchInlineSnapshot('"tw-geqOyW"')
+    expect(frontend({ color: { sm: 'red' } })).toMatchInlineSnapshot('"tw-KGNB_ftuE"')
   })
 })
--- a/packages/core/src/style-decoder.ts
+++ b/packages/core/src/style-decoder.ts
@@ -51,7 +51,10 @@ export class StyleDecoder {
 
     if (hash.className) {
       conds.push(className)
-      result = utility.formatClassName(toHash(conds.join(':')))
+      const stringConds = conds.join(':')
+      const splitConds = stringConds.split('_')
+      const hashConds = splitConds.map(toHash)
+      result = utility.formatClassName(hashConds.join('_'))
     } else {
       conds.push(utility.formatClassName(className))
       result = conds.join(':')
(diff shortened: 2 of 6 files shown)


theMasix:
I'm confused. Is the `cx` supposed to keep the last className of a CSS property and remove the others?
(Related to #1131 and !1195)

It doesn't work as expected!

theMasix:
Yeah `cx` doesn't work as expected

```tsx
  // Logs: text_red text_blue text_green
  console.log(
    cx(
      css({
        color: 'red',
      }),
      css({
        color: 'blue',
      }),
      'text_green',
    ),
  );
```

astahmer:
> Yeah `cx` doesn't work as expected
> 
> ```tsx
>   // Logs: text_red text_blue text_green
>   console.log(
>     cx(
>       css({
>         color: 'red',
>       }),
>       css({
>         color: 'blue',
>       }),
>       'text_green',
>     ),
>   );
> ```

this approach has been discarded in favor of css.raw a while ago

https://github.com/chakra-ui/panda/blob/main/CHANGELOG.md#changed-15
https://panda-css.com/docs/concepts/merging-styles

astahmer:
you might be interested in https://github.com/brendon1555/panda-cx-deduplicator to do it with `cx` only

theMasix:
@astahmer 
This MR resolves this issue:
IMG_20240121_195113

astahmer:
yeah I understood that part, this PR seems fine to me, I just wanted to clarify that `cx` isn't expected to smartly merge things 🙏

---

I'm curious tho, is there anything that prevents you from passing the style object [in those components](https://github.com/chakra-ui/panda/pull/1195#issuecomment-1902683724) rather than directly setting the className ? or is just a preference

theMasix:
@astahmer thanks for understanding and evaluating applied changes.

---

If we want to pass objects down, then we should introduce new prop per component to accept such that object.
Seems weird and kinda non-standard to me.

I always prefer sticking to standard solutions. But it worth talking about in our team. We discuss it.

astahmer:
> If we want to pass objects down, then we should introduce new prop per component to accept such that object.

fwiw: as shown [here](https://panda-css.com/docs/concepts/merging-styles#merging-within-jsx-component), the `css` prop will always automatically be extracted (without the need for `css.raw` wrapping the style object passed)

the `styled` factory already handles this kind of merging, as shown [here](https://play.panda-css.com/JUa2RxiZle)

that being said, I understand wanting to stay close to standard className, again just pointing out that those things can be handled with Panda

theMasix:
We don't use `styled`. But the `css` prop looks great to me.

Thanks for introducing it.

segunadebayo:
This looks interesting @theMasix.

Discussed this PR with @astahmer. I'm a bit cautious about introducing this to the codebase just for the convenience of merging styles.

Panda was designed to merge styles via the `css` prop or `css` function. I recommend using this approach instead of a custom-built tailwind-merge function, which adds extra runtime overhead and JS bundle. This feels like forcing Panda to work like Tailwind or Tailwind Merge. These are two different solutions.

```jsx
const Demo = (props) => {
   return <div {...props} className={css({ bg : "red.300" }, props.css)} />
}
```

The fact that this increases the hash length for every user is quite concerning. If you can find a way to shorten the hash further and keep it between 5-6 characters as we had before, we can consider moving forward with this proposal.

theMasix:
@segunadebayo What's your idea about changing the behavior using a hash option flag?

segunadebayo:
At this point, we can’t accept any new flags. We trying to avoid bloating the config with flags. 

These things are easy to add but ridiculously hard to remove in the future.

theMasix:
I think one of the issues that may occur while using the `css` prop is the thing that writing memorized components become hard.
Imagine a component that has only primitive props. The only prop that has not primitive type is `css`. Then if we want to wrap the component using `memo` function, we should handle the `css` manually and it's a big cons for this solution.

What's your idea about this issue? @segunadebayo

PR #2087, opened 2024-01-26
Add hidden display to preflight

anubra266:
Set `display: none` for hidden elements in `reset` css


packages/generator/src/artifacts/css/reset-css.ts
```diff
@@ -221,6 +221,10 @@ export function generateResetCss(ctx: Context) {
   ${selector}:-moz-focusring {
     outline: auto;
   }
+
+  ${selector}[hidden] {
+    display: none;
```
anubra266, line 226:
```suggestion
    display: none !important;

PR #2088, opened 2024-01-26
feat: config validation

astahmer:
## 📝 Description

Add config validation:

- Check for duplicate between token & semanticTokens names
- Check for duplicate between recipes/patterns/slots names
- Check for token / semanticTokens paths (must end/contain 'value')
- Check for self/circular token references
- Check for missing tokens references
- Check for conditions selectors (must contain '&')
- Check for breakpoints units (must be the same)

> You can set `validate: 'warn'` in your config to only warn about errors or set it to `none` to disable validation
> entirely.

yes if you have errors in your config 😅


packages/config/src/validate-config.ts
```diff
@@ -0,0 +1,250 @@
+import { logger } from '@pandacss/logger'
```
segunadebayo, line 1: Please add tests for all the different scenarios this function covers.

I would also recommend splitting them into dedicated functions so they can be isolated:

- validateTokens
- validateRecipes
- validateBreakpoints
- validateTokenReferences

etc.

Then we can compose them together in this function.

sandbox/vite-ts/panda.config.ts
```diff
@@ -30,11 +30,13 @@ export default defineConfig({
   jsxFactory: 'panda',
   jsxFramework: 'react',
   conditions: {},
+  validation: 'warn',
```
anubra266, line 33: We should add this to the docs

PR #2094, opened 2024-01-27
feat: token/utility/codegen hooks

astahmer:
Closes https://github.com/chakra-ui/panda/pull/2056 https://github.com/chakra-ui/panda/pull/1848

## 📝 Description

Introduce 3 new hooks:

```ts
export interface PandaHooks {
  // ..

  /**
   * Called when the TokenDictionary has been created
   * You can use this hook to add your own tokens to the TokenDictionary
   * You can also override the default `formatTokenName` and `formatCssVar` methods
   */
  'tokens:created': (args: { tokens: TokenDictionary }) => MaybeAsyncReturn
  /**
   * Called when the Utility has been created
   * You can override the default `formatTokenName` and `formatCssVar` methods
   */
  'utility:created': (args: {
    setToHashFn: (fn: (path: string[], toHash: (str: string) => string) => string) => void
  }) => MaybeAsyncReturn
  /**
   * Called right before writing the codegen files to disk.
   * You can use this hook to tweak the codegen files before they are written to disk.
   */
  'codegen:prepare': (args: { artifacts: Artifact[]; changed: ArtifactId[] | undefined }) => MaybeAsyncReturn

  // ..
}
```

no

## 📝 Additional Information

with those hooks, the features from the 2 mentioned PR should be possible without introducing new config options or changing much about Panda internals


packages/core/__tests__/rule-processor.test.ts
```diff
@@ -84,6 +262,7 @@ describe('rule processor', () => {
         },
       },
     })
+    console.timeEnd('css')
```
segunadebayo, line 265: Remove Console.

PR #2101, opened 2024-01-27
feat(hooks): available in presets

astahmer:
## 📝 Description

Allow `config.hooks` to be in `presets`, will sequentially call them

For the `cssgen:done` hook, the return result (if any) of the previous hook call is passed to the next hook (in
`args.content`)

no


segunadebayo (commented):
- The extends keyword
- add codegen:prepare as well

astahmer:
closing in favor of https://github.com/chakra-ui/panda/pull/2186

PR #2112, opened 2024-01-30
fix: 0.29 issue with panda init & color mix on object values

astahmer:
## 📝 Description

Fix an issue (introduced in v0.29) with `panda init` and add an assert on the new `colorMix` utility function

no


packages/core/src/color-mix.ts
```diff
@@ -2,6 +2,7 @@ import type { TransformArgs } from '@pandacss/types'
 
 export const colorMix = (value: string, token: TransformArgs['token']) => {
   if (!value) return { invalid: true, value }
```
segunadebayo, line 4: I believe we can remove this line now. Since the typeof check below will also match

PR #2154, opened 2024-02-06
docs: add more hook docs

anubra266:
Update hooks docs


astahmer:
should we add this use-case ? it's been asked a few times now

https://github.com/chakra-ui/panda/discussions/2155#discussioncomment-8387556

anubra266:
Done @astahmer

PR #2162, opened 2024-02-07
fix: solid jsx factory

### LINKED ISSUE #2156: `styled` fn with components using `as` prop
### Description

When I use the `styled` function from Panda like shown below, to wrap components that also using the `as` prop [Kobalt UI](https://kobalte.dev/) or [Ark UI](https://ark-ui.com) I get the following result and a non functioning component.

`<button  as="function () { [native code] }"></button>`

If the `Button` itself is a `styled` wrapped button I get an `Maximum call stack size exceeded` exception.

```jsx
import * as Dialog from './dialog';
import { styled } from 'styled-system/jsx'

const Button = (props: any) => {
  return <button {...props} />
}

const StyledButton = styled("button")

export const Main = () => {
  return (
    <Dialog.Root>
      <Dialog.Trigger as={Button}>Normal Button</Dialog.Trigger>
      {/* <Dialog.Trigger as={StyledButton}>Maximum call stack size exceeded</Dialog.Trigger> */}
      <Dialog.Backdrop />
      <Dialog.Positioner>
        <Dialog.Content>Panda CSS is awesome</Dialog.Content>
      </Dialog.Positioner>
    </Dialog.Root>
  );
};
```

### Link to Reproduction

https://stackblitz.com/edit/vitejs-vite-tzh3ns?file=src%2Fmain.tsx,src%2Fbutton.tsx,src%2Fdialog.tsx&terminal=dev

segunadebayo:
Closes #2156

## 📝 Description

Fix the issue where styled factory in Solid.js could result in `Maximum call stack exceeded` when composing with another
libraries that use the `as` prop.

## 📝 Additional Information

This was related to how we split props within the factory. The previous approach could lead to extraneous props like "as" being passed to the underlying component.

### DIFF
--- a/.changeset/shy-carrots-teach.md
+++ b/.changeset/shy-carrots-teach.md
@@ -1,0 +1,6 @@
+---
+'@pandacss/generator': patch
+---
+
+Fix issue where styled factory in Solid.js could results in `Maximum call stack exceeded` when composing with another
+library that uses the `as` prop.
--- a/packages/generator/src/artifacts/solid-jsx/jsx.ts
+++ b/packages/generator/src/artifacts/solid-jsx/jsx.ts
@@ -5,63 +5,88 @@ export function generateSolidJsxFactory(ctx: Context) {
   const { componentName, factoryName } = ctx.jsx
   return {
     js: outdent`
-    import { Dynamic } from 'solid-js/web'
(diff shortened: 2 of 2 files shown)


asterikx:
Thank you! This issue has really been driving me crazy.... Hoping that the fix will ship soon and finally unblock me with Park-UI.

PR #2175, opened 2024-02-09
fix: resolve property conflicts and order

astahmer:
Closes https://github.com/chakra-ui/panda/discussions/2165

## 📝 Description

- Sort the longhand/shorthand atomic rules in a deterministic order to prevent property conflicts
- Automatically merge the `base` object in the `css` root styles in the runtime

Ex:

```ts
css({
  padding: '1px',
  paddingTop: '3px',
  paddingBottom: '4px',
})
```

Will now always generate the following css:

```css
@layer utilities {
  .p_1px {
    padding: 1px;
  }

  .pt_3px {
    padding-top: 3px;
  }

  .pb_4px {
    padding-bottom: 4px;
  }
}
```

somewhat, depending on usage


etlovett:
Unfortunately I'm not familiar enough with the codebase to be able to tell from the diff contents, but will this apply the same behavior across merged objects? If we code like `css({ paddingTop: '4px' }, { padding: '6px' })`, for example, will it also sort those properties according to the priority stack, or does it only work within a single object?

packages/shared/src/shorthand-properties.ts
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * ```
+ *
+ * and so the more specific selector would win
+ *
+ * Taken from
+ * @see https://github.com/callstack/linaria/blob/049a4ccb77e29f3628353352db21bd446fa04a2e/packages/atomic/src/processors/helpers/propertyPriority.ts
+ */
+export function getPropertyPriority(property: string) {
+  return longhands.includes(property) ? 2 : 1
```
etlovett, line 135: Is this sufficient complexity, or should there be more priority levels? I notice that in StyleX they define four different layers of priority for normal properties ([code](https://github.com/facebook/stylex/blob/c97421056c819f042271d717eee8ce5c1a4c1669/packages/shared/src/utils/property-priorities.js#L755-L766)), from highest to lowest priority:
1. Longhand physical properties (e.g. `border-bottom-color`)
2. Longhand logical properties (e.g. `border-block-end-color`)
3. Shorthand versions of longhand-only properties (e.g. `border-bottom`)
4. Shorthand versions of shorthand properties (e.g. `border`)

astahmer:
> Unfortunately I'm not familiar enough with the codebase to be able to tell from the diff contents, but will this apply the same behavior across merged objects? If we have code like `css({ paddingTop: '4px' }, { padding: '6px' })`, for example, will it also sort those properties according to the priority stack, or does it only work within a single object?

this PR changes the way the CSS rules are sorted before generating them but after they've all been gathered from the source files
so it's not order dependent, it does not matter where you define those

packages/shared/src/shorthand-properties.ts
astahmer: can you give me an example with `css({ ... })` where the current implementation would be problematic ?

etlovett:
> > Unfortunately I'm not familiar enough with the codebase to be able to tell from the diff contents, but will this apply the same behavior across merged objects? If we have code like `css({ paddingTop: '4px' }, { padding: '6px' })`, for example, will it also sort those properties according to the priority stack, or does it only work within a single object?
> 
> this PR changes the way the CSS rules are sorted before generating them but after they've all been gathered from the source files so it's not order dependent, it does not matter where you define those

Perfect!

PR #2177, opened 2024-02-09
feat: mixed conditions

astahmer:
Closes https://github.com/chakra-ui/panda/discussions/1216

## 📝 Description

Add a way to create config conditions with nested at-rules/selectors

```ts
export default defaultConfig({
  conditions: {
    extend: {
      supportHover: ['@media (hover: hover) and (pointer: fine)', '&:hover'],
    },
  },
})
```

```ts

css({
  _supportHover: {
    color: 'red',
  },
})
```

will generate the following CSS:

```css
@media (hover: hover) and (pointer: fine) {
  &:hover {
    color: red;
  }
}
```

no

### DIFF
--- a/packages/config/src/validation/validate-condition.ts
+++ b/packages/config/src/validation/validate-condition.ts
@@ -19,13 +19,5 @@ export const validateConditions = (conditions: Conditions | undefined, addError:
         addError('conditions', `Selectors should contain the \`&\` character: \`${c}\``)
       }
     })
-
-    const hasAtLeastOneAmperand = condition.some((c) => c.includes('&'))
-    if (!hasAtLeastOneAmperand) {
-      addError(
-        'conditions',
-        `Mixed selectors should contain the \`&\` character in at least one segment: \`${condition.join(', ')}\``,
-      )
-    }
   })
 }
--- a/packages/config/src/validation/validate-condition.ts
+++ b/packages/config/src/validation/validate-condition.ts
@@ -1,12 +1,31 @@
 import type { Conditions } from '@pandacss/types'
 import type { AddError } from '../types'
+import { isString } from '@pandacss/shared'
 
 export const validateConditions = (conditions: Conditions | undefined, addError: AddError) => {
   if (!conditions) return
 
   Object.values(conditions).forEach((condition) => {
-    if (!condition.startsWith('@') && !condition.includes('&')) {
-      addError('conditions', `Selectors should contain the \`&\` character: \`${condition}\``)
+    if (isString(condition)) {
+      if (!condition.startsWith('@') && !condition.includes('&')) {
+        addError('conditions', `Selectors should contain the \`&\` character: \`${condition}\``)
+      }
+
+      return
+    }
+
+    condition.forEach((c) => {
+      if (!c.startsWith('@') && !c.includes('&')) {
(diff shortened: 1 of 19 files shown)


anubra266:
We probably want to update this:
CleanShot 2024-02-09 at 23 36 12@2x

packages/config/src/validation/validate-condition.ts
```diff
@@ -1,12 +1,31 @@
...
+      }
+    })
+
+    const hasAtLeastOneAmperand = condition.some((c) => c.includes('&'))
+    if (!hasAtLeastOneAmperand) {
+      addError(
+        'conditions',
+        `Mixed selectors should contain the \`&\` character in at least one segment: \`${condition.join(', ')}\``,
+      )
     }
```
segunadebayo, line 29: A mixed condition could definitely use two "at-rule" (say @media and @supports). Maybe this is perhaps too strict?

packages/core/src/sort-style-rules.ts
```diff
@@ -1,43 +1,83 @@
...
-  if (a.conditions!.length === b.conditions!.length) {
-    const lastA = a.conditions![a.conditions!.length - 1]
-    const lastB = b.conditions![b.conditions!.length - 1]
 
-    const atRule1 = lastA.params ?? lastA.rawValue
-    const atRule2 = lastB.params ?? lastB.rawValue
+const flatten = (conds: ConditionDetails[]) => conds.flatMap((cond) => (cond.type === 'mixed' ? cond.value : cond))
 
-    if (!atRule1 || !atRule2) return 0
+const compareAtRuleOrMixed = (a: WithConditions, b: WithConditions) => {
```
segunadebayo, line 27: Can you include some tests that cover this new use case? It's a little hard to follow, and I prefer to read the tests instead.

If we could even isolate this bit of the code so we can test it independently, that would be even better.

sandbox/preact-ts/panda.config.ts
```diff
@@ -1,6 +1,11 @@
 import { defineConfig } from '@pandacss/dev'
 
 export default defineConfig({
+  conditions: {
+    mq: '@media (min-width: 640px)',
+    custom: '&[data-custom]',
+    supportHover: ['@media (hover: hover) and (pointer: fine)', '@supports (display: grid)', '&:hover'],
+  },
```
segunadebayo, line 8: Let's move these to a dedicated test somewhere in packages/core so we can be more confident it works in the future.

packages/core/src/sort-style-rules.ts
astahmer: I had done that [here](https://github.com/chakra-ui/panda/blob/ae20e36145e4fddea5a9dd53627389ee93e48743/packages/core/__tests__/rule-processor.test.ts#L1296)

I just added a new one specifically for `compareAtRuleOrMixed` [here](https://github.com/chakra-ui/panda/blob/a0a3c1ee028d85b8ed9390f98cf3111a1b1d153d/packages/core/__tests__/conditions.test.ts#L93)

astahmer: added further improvements to the sorting logic & added a bunch more tests in [this commit](https://github.com/chakra-ui/panda/pull/2177/commits/7d821dbd5f2bc2ce03f8448471e25fc745954bd1) in the rule-processor.test file

PR #2238, opened 2024-02-22
feat: cssVars for generic vars and @property syntax

astahmer:
## 📝 Description

Introduce a new `cssVars` config option to define type-safe
[CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/--*) and custom
[CSS @property](https://developer.mozilla.org/en-US/docs/Web/CSS/@property).

Example:

```ts

export default defineConfig({
  // ...
  cssVars: {
    '--some-color': 'red',
    '--button-color': {
      syntax: '<color>',
      inherits: false,
      initialValue: 'blue',
    },
  },
})
```

> Note: Keys defined in `cssVars` will be available as a value for _every_ utilities, as they're not bound to token
> categories.

```ts
import { css } from '@pandacss/react'

const className = css({
  '--button-color': 'colors.red.300',
  // ^^^^^^^^^^^^  will be suggested

  backgroundColor: 'var(--button-color)',
  //                ^^^^^^^^^^^^^^^^^^  will be suggested
})
```

no


.changeset/plenty-guests-battle.md
```diff
@@ -0,0 +1,43 @@
...
+    },
+  },
+})
+```
+
+> Note: Keys defined in `globalVars` will be available as a value for _every_ utilities, as they're not bound to token
+> categories.
+
+```ts
+import { css } from '@pandacss/react'
```
segunadebayo, line 34:
```suggestion
import { css } from '@pandacss/dev'
```

packages/generator/src/artifacts/css/token-css.ts
segunadebayo, line 1: I think the global css should be one responsible for generating the css, not tokens

astahmer: then maybe we should rename this to `tokensRoot` ?
image

PR #2269, opened 2024-02-26
feat: utility group

astahmer:
Add a `group` to every utility in the `@pandacss/preset-base`, this helps Panda tooling organize utilities.


segunadebayo:
Let’s update the name of groups to singular format. For example,

List
Flex Layout
Grid Layout
Transition
Transform
Effect
Other
Border
Interactivity
Scroll

- Inconsistent grouping for border* and outline* properties 
- Move divide* properties to Border
- Touch action and user select
- Over scroll props to Scroll

PR #2313, opened 2024-03-06
fix: _active condition was triggering when data-active=false

6qat:
There is a problem with the "_active" condition. If the component has the attribute data-active=false, it is triggered. Just the presence of it triggers the condition, regardless of its value. I had to override it to make it work properly.

This correction probably should be made in many other default conditions later.

Closes # 

## 📝 Description

> Add a brief description

> Please describe the current behavior that you are modifying

## 🚀 New behavior

> Please describe the behavior or changes this PR adds

### DIFF
--- a/.changeset/_active-condition-mistaken.md
+++ b/.changeset/_active-condition-mistaken.md
@@ -1,0 +1,7 @@
+---
+"@pandacss/preset-base": patch
+---
+
+_active condition was triggering when data-active=false
+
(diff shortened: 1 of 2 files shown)


segunadebayo:
Hi @6qat,

This is not a bug in the conditions. The expectation is that your code should not rely on true/false for the data attributes.

You can solve this like so:

```tsx
<button data-active={isActive? "": undefined} />
```

instead of 

```tsx
<button data-active={isActive} />
```

anubra266:
@6qat to add more to this. It's invalid semantic html to have `data-active=false` you either have a data attribute or not on your element, it's not supposed to have true or false values. 
So the way to ensure this is what @segunadebayo has suggested above.

PR #2323, opened 2024-03-07
fix: validation issue when referencing a token with DEFAULT

astahmer:
## 📝 Description

follow-up to https://github.com/chakra-ui/panda/pull/2289 where I left a typo
makes token referencing path with DEFAULT not throw for the validation check


anubra266:
@astahmer Prettier fails

PR #2353, opened 2024-03-10
fix: allow css files in node_modules/.embroider

IgnaceMaes:
## 📝 Description

Fixes the PostCSS plugin to be compatible with [Embroider](https://github.com/embroider-build/embroider).

Embroider virtualizes app import files in `/node_modules/.embroider/`. Currently this gets skipped as it's inside `/node_modules/`.

See also: https://github.com/IgnaceMaes/ember-pandacss-demo

CSS files inside of `/node_modules/` are skipped.

## 🚀 New behavior

CSS files inside of `/node_modules/` are skipped, except if it's inside of `/node_modules/.embroider/`.

No

## 📝 Additional Information

❤️

### DIFF
--- a/packages/postcss/src/index.ts
+++ b/packages/postcss/src/index.ts
@@ -71,6 +71,8 @@ pandacss.postcss = true
 export default pandacss
 
 const nodeModulesRegex = /node_modules/
+// Embroider virtualizes files inside of `/node_modules/.embroider/rewritten-app`
+const nodeModulesEmbroiderRegex = /node_modules\/.embroider/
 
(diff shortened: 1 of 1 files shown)


anubra266:
Thanks for this @IgnaceMaes. I'll be closing in favour of #2354, because we'd like to avoid hardcoding things like this. 
We'd rather expose a way for you to configure this and any other related usecases in the future.

IgnaceMaes:
Agree that's the best way forward.

Thanks for taking a look.

PR #2419, opened 2024-03-22
BREAKING: Pluralise token categories

anubra266:
- Pluralise tokens
- Refactor tokens usage in repo


segunadebayo:
Thanks for this PR @anubra266 

Discussed with Alex about this. We decided to keep the token name as is for now.

PR #2452, opened 2024-04-01
fix(studio): display escape hatch values in textStyles

astahmer:
## 📝 Description

- Handle displaying values using the `[xxx]` escape-hatch syntax for `textStyles` in the studio
- Display an empty state when there's no token in a specific token page in the studio

no

## 📝 Additional Information

- Add `deepResolveReference` in TokenDictionary, helpful to get the raw value from a semantic token by recursively traversing the token references.
- Added some exports in the `@pandacss/token-dictionary` package, mostly useful when building tooling around Panda (Prettier/ESLint/VSCode plugin etc)

### DIFF
--- a/packages/studio/astro.config.mjs
+++ b/packages/studio/astro.config.mjs
@@ -1,9 +1,8 @@
-import react from '@astrojs/react'
 import studio from '@pandacss/astro-plugin-studio'
 import { defineConfig } from 'astro/config'
 
 // https://astro.build/config
 export default defineConfig({
   devToolbar: { enabled: true },
-  integrations: [react(), studio()],
+  integrations: [studio()],
 })
--- a/packages/studio/src/components/font-tokens.tsx
+++ b/packages/studio/src/components/font-tokens.tsx
@@ -5,6 +5,8 @@ import { TokenContent } from '../components/token-content'
 import { TokenGroup } from '../components/token-group'
 import { Input, Textarea } from './input'
 import { StickyTop } from './sticky-top'
+import { EmptyState } from './empty-state'
+import { TypographyIcon, XMarkIcon } from './icons'
 
 interface FontTokensProps {
   text?: string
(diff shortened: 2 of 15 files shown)


packages/studio/src/components/font-tokens.tsx
```diff
@@ -23,6 +25,14 @@ export default function FontTokens(props: FontTokensProps) {
     setText(event.target.value)
   }
 
+  if (fontTokens.length === 0) {
+    return (
+      <EmptyState title="No Tokens" icon={<TypographyIcon />}>
+        The panda config does not contain any `{token}` tokens
```
segunadebayo, line 31:
```suggestion
        The panda config does not contain any font tokens
```

packages/studio/astro.config.mjs
```diff
@@ -1,9 +1,8 @@
-import react from '@astrojs/react'
```
segunadebayo, line 1: Removing this will cause Astro to throw a "React" is not defined error in the users' project. Esp when they don't use react.

packages/studio/src/components/font-tokens.tsx
astahmer: no I think we should keep it as is, as it depends on which page is rendering this component
https://github.com/chakra-ui/panda/blob/97e1152524af9a1ccd7a6085599919055e07fb63/packages/studio/src/pages/font-sizes.astro#L12
https://github.com/chakra-ui/panda/blob/97e1152524af9a1ccd7a6085599919055e07fb63/packages/studio/src/pages/font-weights.astro#L12

segunadebayo: Oh I see. Then the strong interpolation needs to be updated.

`${token}`

packages/studio/astro.config.mjs
segunadebayo: Yes. I think we can remove that one. It’s not used

PR #2466, opened 2024-04-03
refactor: make `WithImportant<T>` more performant

### LINKED ISSUE #2457: `WithImportant<T>` makes TypeScript extremely slow
### Description

When diagnosing the codebase with [svelte-check](https://github.com/sveltejs/language-tools/tree/master/packages/svelte-check) (which has the type checking feature), `WithImportant<T>` that is defined part of `WithEscapeHatch<T>` makes the whole process extremely slow.

With `WithImportant<T>`:
```ts
...
 export type WithEscapeHatch<T> =
   | T
   | `[${string}]`
   | (T extends string ? WithColorOpacityModifier<string> | WithImportant<T> : T);
...
```

`svelte-check` takes 4m58s to complete.

Without it:
```ts
 export type WithEscapeHatch<T> =
   | T
   | `[${string}]`
   | (T extends string ? WithColorOpacityModifier<string> : T);
```

`svelte-check` takes 12s to complete.

---

I guess the type definition of `WithImportant<T>` somehow makes typescript compiler unhappy.

### Link to Reproduction

none

### Steps to reproduce

_No response_

### JS Framework

SvelteKit (TS)

### Panda CSS Version

0.37.0

### Browser

_No response_

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

devunt:
Closes #2457 

## 📝 Description

Make `WithImportant<T>` more performant and typescript happy by updating the type signature to use branded type, while keeping such tokens valid and also not appearing in autocompletions to prevent them from polluting autocompletion result (which is the current behavior)

Current type signaure of `WithImportant<T>` somehow makes typescript compiler unhappy and slows down both code autocompletion and linting. It was also allows some invalid tokens such as `gray.500!abcd` by accepting any arbitrary string after the important token.

## 🚀 New behavior

New type signaure of `WithImportant<T>` is more performant and and stricter. On using `svelte-check` in large codebase was taking over 5 minutes but now it takes under 20 seconds. VSCode autocompletion is also nearly instant, instead of taking ~5 seconds to show.

Technically, yes. But it is negligible.


astahmer:
thank you for the PR ! 🙏 
I've checked out the branch locally and it seems it breaks a few `sandbox/codegen` tests (generated files tested at runtime in the DOM)

image

I've noticed that `vitest` doesn't report those, because vitest `typecheck` was not enabled, sorry about that. I've added those checks in https://github.com/chakra-ui/panda/pull/2468 and rebased your branch based on my PR so that you can benefit from the working `tsc` tests

image

---
steps to reproduce:
```sh
pnpm i && cd sandbox/codegen
bun run ./cli.ts codegen 
bun run ./cli.ts test 
```

> you can use `tsx` instead of bun like so `tsx ./cli.ts codegen`

you can run a specific test with `MODE=scenario pnpm vitest` where `scenario` is one of those
https://github.com/chakra-ui/panda/blob/ab32d1d798c353ce4793f01b1e5cae3407bc209e/sandbox/codegen/cli.ts#L7-L19

alternatively with `bun run ./cli.ts test {scenario}`
https://github.com/chakra-ui/panda/blob/d30b17371c2a8b65df6c971389c86e0448bb3586/sandbox/codegen/README.md#L2-L7

you can update the snapshots with `-u` and watch with `-w` (just like vitest)

PR #2491, opened 2024-04-08
feat: support deprecations (token, utilities, patterns, recipes)

segunadebayo:
## 📝 Description

Add support for deprecating tokens, utilities, patterns and config recipes.

Set the `deprecated` property to `true` to enable deprecation warnings. Alternatively, you can also set it to a string
to provide a custom migration message.

**Deprecating a utility**

```js
defineConfig({
  utilities: {
    ta: {
      deprecated: true,
      transform(value) {
        return { textAlign: value }
      },
    },
  },
})
```

**Deprecating a token**

```js
defineConfig({
  theme: {
    tokens: {
      spacing: {
        lg: { value: '8px', deprecated: 'use `8` instead' },
      },
    },
  },
})
```

**Deprecating a pattern**

```js
defineConfig({
  patterns: {
    customStack: {
      deprecated: true,
    },
  },
})
```

**Deprecating a recipe**

```js
defineConfig({
  theme: {
    recipes: {
      btn: {
        deprecated: 'will be removed in v2.0',
      },
    },
  },
})
```

### ESLint Plugin

These deprecation warnings will translated into the ESLint plugin as a `no-deprecated` rule.

```json
{
  "rules": {
    "no-deprecated": "warn"
  }
}
```

In the next release of the ESLint plugin, you will get a warning when using deprecated tokens or utilities.


astahmer (approved):

anubra266 (commented):
1. It's not accepting the string version

CleanShot 2024-04-08 at 22 42 22@2x

2. It throws validation errors (It's taking `deprecated` key as a nested token

CleanShot 2024-04-08 at 22 44 26@2x

3. We could add tests that have the deprecated jsdoc in their snapshot output

anubra266 (commented):
We're not keeping track of deprecated patterns in `packages/core/src/patterns.ts`
So we can't currently detect if a patterns is runtime.
Expected possibility:
```js
PandaContext.patterns.isDeprecated(patternName)
```

segunadebayo:
Good catch @anubra266. Resolved most of the comments

PR #2494, opened 2024-04-10
chore(docs): remove duplicate tags

38elements:
## 📝 Description

`description`, `og:title`, `og:description`, `og:image` and `og:url` are duplicated in web page. These tags are provided by [metadata](https://github.com/chakra-ui/panda/blob/bc09d893a63ab2b40f6fe026ab3b7bd62dad289d/website/app/layout.tsx#L15) and theme. This removed the tags  provided by the theme.

No

### DIFF
--- a/website/theme.config.tsx
+++ b/website/theme.config.tsx
@@ -8,7 +8,6 @@ import {
   FileTree,
   Tab,
   Tabs,
-  useConfig
 } from './src'
 import { Steps } from './src/mdx/steps'
 import { css } from './styled-system/css'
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks for this PR, @38elements. The app metadata only affects the homepage, not the docs page.

I wonder if there's a way to know (maybe from process.env) that Next.js is rendering the app/pages dir. That way we can conditionally render those metadata

38elements:
Thank you for reply. I confirmed  top page (`/`) and docs page (`/docs`) and learn page (`/learn`) by its sources in local. These page sources work normally.

segunadebayo:
I'm unsure how to solve this @ 38elements.

Let's keep this until I get to a solution that ensures we don't break the open graph of other pages.

PR #2497, opened 2024-04-10
perf: use lru cache

segunadebayo:
## 📝 Description

This is mostly a perf optimization to prevent memory from growing infinitely in a long-running + large project.

We need to mark the hot cache regions in the code path and add the maximum cache size.

Very rarely, but it can sometimes consume a lot of memory.

## 🚀 New behavior

Recomputes after 500 entries

No.


astahmer:
we probably should add this to `StyleDecoder` / `StyleEncoder` too

packages/core/__tests__/utility.test.ts
```diff
@@ -2787,7 +2787,7 @@ describe('Utility', () => {
     expect(utility.getTypes()).toMatchInlineSnapshot(`Map {}`)
   })
 
-  test('skip type:Token[category] when category values is empty', () => {
+  test.only('skip type:Token[category] when category values is empty', () => {
```
segunadebayo: Let's remove the test.only

PR #2498, opened 2024-04-11
fix(tokens): deprecated token errors

anubra266:
Fixed deprecated token errors


segunadebayo:
Thanks for looking into this. 

I've fixed the underlying issue instead of in the main branch.

PR #2503, opened 2024-04-12
feat: strictTokens: "extends" to allow for known values of CssPropert…

astahmer:
…ies in addition to tokens values

Closes # 

## 📝 Description

Introduce a new `extends` value for `strictTokens`

Setting this in your config will let you use the known values of a CSS property (in addition to the values defined in
your theme).

Example:

```ts
// panda.config.js
export default defineConfig({
  //...
  strictTokens: 'extends',
})

// app.tsx
import { css } from './styled-system/css'

css({ color: 'blue' }) // ✅ will work as `blue` is a native value for the CSS color property
css({ color: 'blue.300' }) // ✅ will work as `blue.300` is defined in your theme by default through the built-in `@pandacss/preset-panda`

// the global CSS values are also available
css({ color: 'initial' }) // ✅
css({ color: 'inherit' }) // ✅
css({ color: 'currentcolor' }) // ✅
css({ color: 'revert' }) // ✅
css({ color: 'revert-layer' }) // ✅
css({ color: 'unset' }) // ✅

// or another example with the width property
css({ width: 'min-content' }) // ✅
css({ width: 'max-content' }) // ✅

// and anything else will still not be allowed
css({ color: '#FFF' }) // ❌ will throw a type error
css({ width: '123px' }) // ❌ will throw a type error
```

no

## 📝 Additional Information

https://github.com/chakra-ui/panda/discussions/2159#discussioncomment-9096306

### DIFF
--- a/.changeset/sweet-cups-sit.md
+++ b/.changeset/sweet-cups-sit.md
@@ -11,6 +11,14 @@ your theme).
 Example:
 
 ```ts
+// panda.config.js
+import { defineConfig } from '@pandacss/dev'
+export default defineConfig({
+  //...
+  strictTokens: 'extends',
+})
+
+// app.tsx
 import { css } from './styled-system/css'
 
 css({ color: 'blue' }) // ✅ will work as `blue` is a native value for the CSS color property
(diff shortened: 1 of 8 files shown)


.changeset/sweet-cups-sit.md
```diff
@@ -0,0 +1,30 @@
...
+'@pandacss/types': patch
+---
+
+Introduce a new `extends` value for `strictTokens`
+
+Setting this in your config will let you use the known values of a CSS property (in addition to the values defined in
+your theme).
+
+Example:
+
```
anubra266, line 12:
```ts
// panda.config.js

export default defineConfig({
  //...
  strictTokens: 'extends',
})
```

segunadebayo:
After discussing this, we agreed to use the escape hatch syntax for these cases.

PR #2514, opened 2024-04-17
docs: clean up fontsource guide

ayuhito:
## 📝 Description

This cleans up the Fontsource guide a little and fixes the incorrect examples. Additionally, I recommend pushing developers to use variable fonts for more flexibility (and slowly educate it to new developers as the new norm which is also what Google Fonts is doing).

No

## 📝 Additional Information

I also maintain Fontsource and was wondering if you wanted to remove the [Chakra UI guide](https://fontsource.org/docs/guides/angular) from the sidebar and want to replace that with PandaCSS? If I understand correctly, Chakra UI is no longer the focus, right? I can make that change for you if that is ideal.

### DIFF
--- a/website/pages/docs/guides/fonts.md
+++ b/website/pages/docs/guides/fonts.md
@@ -74,29 +74,29 @@ export default function App({ Component, pageProps }) {
 }
 ```
 
-### Font Source
+### Fontsource
 
-Fontsource provides a straightforward process for installing fonts into your web application.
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks for this PR, @ayuhito

Chakra UI is still in focus, and we'll launch v3 shortly. We don't intend to replace Chakra UI with Panda in your guide; instead we would love to have an additional guide for Panda in your docs.

PR #2522, opened 2024-04-23
fix(postcss): set postcss as peer dependency

### LINKED ISSUE #2520: PandaCSS' postcss dependency conflicts with vite's postcss dependency
### Description

When I add the postcss plugin in my `vite.config.ts` file I get a TS error: 
image

After investigating the issue I noticed I have two postcss versions: 8.4.35 (PandaCSS' required version) and 8.4.38 (Vite's version).

Since the postcss plugin is just using the postcss type definition, would it be possible to use a peer dependency instead?

Thanks!

### Link to Reproduction

https://github.com/rcoedo/pandacss-postcss-bug

### Steps to reproduce

Open `vite.config.ts` and there will be a typescript error like the one mentioned.

### JS Framework

React

### Panda CSS Version

0.37.2

### Browser

_No response_

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

rcoedo:
Closes #2520 

## 📝 Description

This PR turns the `postcss` dependency into a peer dependency.

`postcss` is being installed as a dependency with a fixed version.

## 🚀 New behavior

`postcss` will be a peer dependency. End users will be the ones providing an appropriate postcss version.

Potentially, if users were using a postcss version lower than 8.4.x. However, if they were doing that it's possible they were also experiencing some weird behaviour.

## 📝 Additional Information

This is my first contribution to the repository so I'm probably missing some context. I'd appreciate a thorough review from a more experienced contributor!

I hope this helps 😄

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -36,6 +36,9 @@
   "publishConfig": {
     "access": "public"
   },
+  "peerDependencies": {
+    "postcss": "^8.4.x"
+  },
   "dependencies": {
     "@csstools/postcss-cascade-layers": "4.0.4",
     "@pandacss/is-valid-prop": "workspace:^",
@@ -48,7 +51,6 @@
     "lightningcss": "1.23.0",
     "lodash.merge": "4.6.2",
     "outdent": "0.8.0",
-    "postcss": "8.4.35",
     "postcss-discard-duplicates": "6.0.1",
     "postcss-discard-empty": "6.0.1",
     "postcss-merge-rules": "6.0.3",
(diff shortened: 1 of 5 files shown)


segunadebayo:
I appreciate the swift PR, @rcoedo. Thank you

I spoke with the team, and we don't want to go in this direction. To use Panda, we want to keep it just the `@pandacss/dev` dependency.

I recommend using the `postcss.config.js` file, rather than explicitly adding it to Vite yourself which might run into type mismatch occasionally.

It might be a little inconvenient on your end, but it should be okay for most cases.

Thanks for understanding

PR #2607, opened 2024-05-19
refactor: make codegen result compatible with older browsers.

theMasix:
## 📝 Description

The result of `codegen` is a file that has enough browser compatibility except it has some `replaceAll` methods in `helpers.mjs` which prevents the codegen results from great browser compatibility

> NOTE: with this minor improvements, the codegen result now is compatible with `iOS <= 12`

It has some `replaceAll` methods in two different shared utility.

## 🚀 New behavior

I've replaced the two `replaceAll` methods with pure `replace`, `split`, and `join` methods.

NO

### DIFF
--- a/packages/shared/src/classname.ts
+++ b/packages/shared/src/classname.ts
@@ -36,7 +36,7 @@ const fallbackCondition: NonNullable<CreateCssContext['conditions']> = {
   breakpoints: { keys: [] },
 }
 
-const sanitize = (value: any) => (typeof value === 'string' ? value.replaceAll(/[\n\s]+/g, ' ') : value)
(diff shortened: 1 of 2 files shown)


segunadebayo:
@theMasix Thanks for this. Could you add some tests for this change?

theMasix:
> @theMasix Thanks for this. Could you add some tests for this change?

Is there any test for these methods already?

segunadebayo:
After thinking about this, I'd recommend installing a polyfill in your app instead.

- Install `core-js`
- `import 'core-js/features/string/replace-all`

Alternatively, you can run the `codegen:prepare` hook to modify these methods.

Thanks for understanding.

PR #2612, opened 2024-05-20
chore: rm CacheMap - fix HMR in some setups

astahmer:
Closes https://github.com/chakra-ui/panda/discussions/2608

## 📝 Description

Fix a cache issue that leads to HMR growing slower in some cases

no


Pinpickle:
@segunadebayo this PR broke my Panda setup for some reason. I mentioned it on Discord but nobody got back to me.

It's really hard to make a reproduction. Here's what I've noticed when using Vite with the postcss plugin. When I first load my app, the CSS is loaded fine. Then, when I save any file in my project, all of the CSS except for the styles in that file goes away. Then, it incrementally adds styles as I go and save other files.

Reverting this change (turning the Map constructors back into CacheMap constructors) fixes the issue somehow.

PR #2646, opened 2024-06-02
refactor(artifacts): ArtifactMap

astahmer:
it mostly makes it way easier to track (config) dependencies (= which `config.xxx` path change on HMR should trigger a new codegen for a given file) for each artifacts

typings got improved, `map.getFile("xxx")` will suggest any `PossibleArtifact` (= an `ArtifactFileId` that might be generated given a combination of config flags like `jsxFramework`, `recipes`, `syntax` ..) 

when an `ArtifactFile` is generated, any other `ArtifactFileId` used in the `imports` (or `importTypes`) will also be used to transitively generated its (file) dependencies

Ideally, I'd like to find a way to make this `ArtifactMap` available in some `hook` so that any user could use it as a (very minimal) template engine with the `code` method and `computed` arguments of each `ArtifactFile` but this might prove a challenge due to the `@pandacss/types` package not knowing about the `Context` from `@pandacss/core`


segunadebayo:
I am closing this for now. If we ever need it, we'll bring it back up.

PR #2681, opened 2024-06-20
feat: allow custom TokenCategory in tokens/semanticTokens

astahmer:
## 📝 Description

Slightly change the typings so that you can use any string as a `TokenCategory` (colors, spacing, etc).

```ts

export default defineConfig({
  // ...
  theme: {
    extend: {
      tokens: {
        filters: { // ✅ this is now allowed !
          blurry: {
            value: 'blur(5px)',
          },
        },
      },
  },
   utilities: {
    extend: {
      filter: {
        values: 'filters', // ✅ you can reference your own TokenCategory here
      },
    },
  },
})
```

so that you can use it in your app like this:

```tsx

export const App = () => {
  return (
    <div
      className={css({
        filter: 'blurry', // ✅ this will be suggested by the autocomplete
      })}
    >
      this text will be blurry
    </div>
  )
}
```

no

### DIFF
--- a/.changeset/brown-ghosts-juggle.md
+++ b/.changeset/brown-ghosts-juggle.md
@@ -14,14 +14,16 @@ export default defineConfig({
   theme: {
     extend: {
       tokens: {
-        filters: { // ✅ this is now allowed !
+        filters: {
+          // ✅ this is now allowed !
           blurry: {
             value: 'blur(5px)',
           },
         },
       },
+    },
   },
-   utilities: {
+  utilities: {
     extend: {
       filter: {
         values: 'filters', // ✅ you can reference your own TokenCategory here
--- a/.changeset/brown-ghosts-juggle.md
+++ b/.changeset/brown-ghosts-juggle.md
@@ -1,0 +1,50 @@
+---
+'@pandacss/token-dictionary': patch
+'@pandacss/parser': patch
+'@pandacss/types': patch
+---
+
+Slightly change the typings so that you can use any string as a `TokenCategory` (colors, spacing, etc).
+
+```ts
+import { defineConfig } from '@pandacss/dev'
+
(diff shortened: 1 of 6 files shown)


.changeset/brown-ghosts-juggle.md
```diff
@@ -0,0 +1,50 @@
...
+  // ...
+  theme: {
+    extend: {
+      tokens: {
+        filters: { // ✅ this is now allowed !
+          blurry: {
+            value: 'blur(5px)',
+          },
+        },
+      },
```
anubra266, line 22:
```suggestion
      },
  },
```

packages/types/src/tokens.ts
```diff
@@ -91,12 +91,19 @@ export interface TokenDataTypes {
   containerNames: string
 }
 
+export type DefaultTokenCategory = keyof TokenDataTypes
+export type TokenCategory = DefaultTokenCategory | (string & {})
+
+type UnknownTokenTypeValue = string | number | Record<string, any>
```
anubra266, line 97:
```suggestion
type UnknownTokenTypeValue = string | number
```

astahmer: I initially added this & Array<any> cause token types like `borders` / `gradients` / `shadows` allow those, should we ignore those use-cases ?

anubra266: We'd definitely need it for the inbuilt tokens. The idea was to not allow it for generic tokens. If generic tokens accept `Record` there'd have to be a way to transform them to their string value. E.g. An object for custom token `dropShadow` would need to be transformed just like we transform `shadows` internally.

astahmer: oh, makes sense, thanks ! 🙏

PR #2692, opened 2024-06-24
refactor: Ensure all utility class names are unique

Jonas-C:
## 📝 Description

The current class names generated by utilities are not unique. This PR ensures that all utility class names are unique, allowing for a potential inlining of something like https://github.com/brendon1555/panda-cx-deduplicator. In any case, having unique class names for each utility is a good thing. 

... Yes? If users are relying on the existing class names. Should probably be a minor (0.x.0) bump of preset-base.

## 📝 Additional Information

* I don't really know if you have a predefined pattern for defining class names, so what I did is a little all over the place. Let me know if you do (or if you don't and want to work something out!) and I'll fix it. Nonetheless, I added two tests to ensure that all utilities have class names and that all class names are unique.
* I added a TODO-comment for preset-open-props. I don't really know how it's used. `opacity` is duplicated there.
* I omitted changelogs because I'm unsure how you want to handle them. Should this only be a minor bump for preset-panda?


packages/preset-open-props/src/utilities.ts
```diff
@@ -21,6 +21,7 @@ export const utilities: UtilityConfig = {
     },
   },
   filter: {
+    // TODO: Does this need to be unique?
```
anubra266, line 24: It's safe to assume, open props is not used with panda preset, so this doesn't need to use a different classname from filter in panda preset.

anubra266:
@Jonas-C What do you mean opacity is duplicated

packages/preset-open-props/src/utilities.ts
Jonas-C: These changes do not apply to the panda preset, but rather the base preset. But if only the last one is resolved I guess it's not a problem?

Jonas-C:
> @Jonas-C What do you mean opacity is duplicated

@anubra266 My bad, I meant filter. It's defined as `className: 'filter'` in both preset-base and preset-open-props.

anubra266:
@Jonas-C 
1. In that case, it's fine to have them with same classnames, since they're different presets. If you use both presets, only one of them will remain when the presets are merged in the config
2. Let's put those new tests in the preset-base package since that's the only thing it covers, it doesn't any relationship with the core package, There are currently no tests in the preset-base package, but it's fine to add it there

Jonas-C:
@anubra266 Done! Let me know if you want me to change some of the class names:)

segunadebayo:
Made some changes to the class name convention. I will review it tomorrow and get it merged soon.

We should have unique class names across the board. Thanks for adding those tests as well.

PR #2694, opened 2024-06-24
refactor(BREAKING): rm config.eject & automatic import of built-in pr…

astahmer:
…esets

Closes # 

## 📝 Description

- BREAKING: The built-in presets `@pandacss/preset-base` and `@pandacss/preset-panda` are no longer automatically
  included, if you are not using `presets` or `eject` in your config you can add them manually as shown below:

```diff
// panda.config.ts
import { defineConfig } from '@pandacss/dev'

export default defineConfig({
+  presets: ['@pandacss/preset-base', '@pandacss/preset-panda'],
})
```

This allows you to control more explicitely which presets are included in your project.

- BREAKING: The `config.eject` option has been removed. Simply remove it from your config and the result will be the
  same.

```diff
// panda.config.ts
import { defineConfig } from '@pandacss/dev'

export default defineConfig({
-  eject: true,
})
```

yes


packages/node/src/setup-config.ts
```diff
@@ -42,8 +42,12 @@ export async function setupConfig(cwd: string, opts: SetupOptions = {}) {
   } else {
     const content = outdent`
 import { defineConfig } from "@pandacss/dev"
+import presetBase from "@pandacss/preset-base"
+import presetPanda from "@pandacss/preset-panda"
 
 export default defineConfig({
+    // The set of reusable and shareable configuration presets.
+    presets: [presetBase, presetPanda],
```
segunadebayo, line 50: Let's use the string version of the preset so users don't have to install these packages.

`presets: ['@pandacss/preset-base', '@pandacss/preset-panda'],`

PR #2701, opened 2024-06-27
feat: support `animationStyle` composition

segunadebayo:
## 📝 Description

This is a spin-off from my work on Chakra. Let's introduce a new "mixin" called motion styles. The idea is to pair them with text styles and layer styles to create this three-part mixin that can make your styles a lot cleaner.

Motion styles focus solely on animations, allowing you to orchestrate animation properties. 

Here's a good example of this:

```jsx
import { defineMotionStyles } from "@pandacss/dev"

export const motionStyles = defineMotionStyles({
  "slide-fade-in": {
    value: {
      transformOrigin: "var(--transform-origin)",
      animationDuration: "fast",
      "&[data-placement^=top]": {
        animationName: "slide-from-top, fade-in",
      },
      "&[data-placement^=bottom]": {
        animationName: "slide-from-bottom, fade-in",
      },
      "&[data-placement^=left]": {
        animationName: "slide-from-left, fade-in",
      },
      "&[data-placement^=right]": {
        animationName: "slide-from-right, fade-in",
      },
    },
  },
})
```

With that defined, I can use it in my recipe or css like so:

```js
export const popoverSlotRecipe = defineSlotRecipe({
  slots: anatomy.keys(),
  base: {
    content: {
      _open: {
        motionStyle: "scale-fade-in",
      },
      _closed: {
        motionStyle: "scale-fade-out",
      },
    },
  },
})

```

This feature will allow us to lean in towards CSS for animations rather than JS. Composing animation names is a powerful feature we should encourage consumers to use.


caseybaggz:
LFG! I'm ready to use this 🔥

PR #2790, opened 2024-08-08
chore: switch to package-manager-detector to reduce dependencies

benmccann:
## 📝 Description

Drops 24 dependencies

https://npmgraph.js.org/?q=preferred-pm - 24 dependencies

## 🚀 New behavior

https://npmgraph.js.org/?q=package-manager-detector - 0 dependencies

No


packages/node/src/setup-config.ts
```diff
@@ -28,8 +28,8 @@ export async function setupConfig(cwd: string, opts: SetupOptions = {}) {
     }
   }
 
-  const pmResult = await getPackageManager(cwd)
-  const pm = pmResult?.name ?? 'npm'
+  const pmResult = await detect({ cwd })
+  const pm = (pmResult?.agent ?? 'npm').split('@')[0]
```
segunadebayo, line 32:
```suggestion
  const pm = (pmResult ?? 'npm').split('@')[0]
```

I believe `detect` returns a string, so accessing `.agent` will return `undefined,` right?

benmccann: It returns an object with `agent` and `version` properties, so I believe what I have is correct

Screenshot from 2024-08-09 12-57-23

PR #2841, opened 2024-09-06
perf(shared): simplify assertion in walkObject

Jonas-C:
Trying my hand at some low-hanging perf improvements while getting to know the code base!

## 📝 Description

Reduce the amount of checks needed within `walkAbout`

`isObject` checks if `value` is an actual object, not null and not an array. `Array.isArray` checks if the value is an array.

## 🚀 New behavior

`isObjectOrArray` checks that `value` is an object and not null.

No, unless you're relying deeply on panda internals.

## 📝 Additional Information

I think I need to update a generator file for styled-system to work correctly, is that right?

Do you want changesets for stuff like this?


segunadebayo:
Thanks for this. Please run the dev command to sync the shared json files

Jonas-C:
Rebased it into a single commit, seeing as both commits are kind of dependent on the generated code

segunadebayo:
@Jonas-C Could you resolve this one as well?

PR #2846, opened 2024-09-07
perf(artifacts): do not construct the condition regex within `isCondition` function body

Jonas-C:
## 📝 Description

More low-hanging fruit. Just moving the regex creation out of the function body.

No


packages/studio/styled-system/helpers.mjs
Jonas-C, line 1: These changes are automatically added when committing. I'll rebase this away if my other PR is merged

PR #2847, opened 2024-09-07
perf(shared): improve uniq performance

Jonas-C:
## 📝 Description

Creating a new set and a new array for every iteration of the `reduce` function is unnecessary. Instead, we can initialize the reduce with an empty set and continuously add items to it, before ultimately turning it into an array before returning it. According to tinybench, this nearly doubles the throughput of `uniq`.

No


segunadebayo:
@Jonas-C could you share the tiny bench results? A screenshot or url would suffice

Jonas-C:
image
https://gist.github.com/Jonas-C/a62360e50a59148b7f883b726e8b0277

segunadebayo:
@Jonas-C could you rebase to resolve the conflict. Thanks

PR #2848, opened 2024-09-07
perf(shared): Store longhand properties as a map instead of as an array

Jonas-C:
## 📝 Description

This should decrease the initialization time of `longhands` (marginally). The most important thing is that tinybench reports something like an 800% increase in throughput for this specific function.

No

## 📝 Additional Information
I'll rebase this later


segunadebayo:
@Jonas-C could you share the tiny bench results? A screenshot or url would suffice

Jonas-C:
image
https://gist.github.com/Jonas-C/eee842a14962137a1e1f69c0bef948d7

PR #2855, opened 2024-09-09
fix: flex basis tokens

adbutterfield:
## 📝 Description

Since [flex-basis](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis) specifies a width, it should use `sizes` tokens.

This PR updates the type of the `Flex` utility to allow using `sizes` tokens instead of `spacing` tokens.

Type does not allow using `sizes` tokens for `basis` prop on `Flex` utility, but it does allow for using `spacing` tokens.

## 🚀 New behavior

Type now allows using `sizes` tokens for `basis` prop on `Flex` utility, and does not allow using `spacing` tokens.

Technically no? You could always `@ts-ignore` any place where you're currently passing a `spacing` token to `basis`.


segunadebayo:
Thanks @adbutterfield

PR #2907, opened 2024-10-06
feat: add variants precedence in config recipes

cs-clarence:
## 📝 Description

adds `variantsPrecedence` option to `defineRecipe` and `defineSlotRecipe` to specify the order of how the variants' styles are generated in the CSS.

https://play.panda-css.com/MjjHfvheHA

Not applied for atomic recipes because the order of variants specified in the recipe usage calls already determines which styles are applied.

```ts
const atomicTest = cva({
  variants: {
    active: {
      true: {
        bg: "red",
      },
    },
    disabled: {
      true: {
        bg: "blue",
      },
    },
  },
})

const cvaT1 = atomicTest({
  active: true,
  disabled: true,
})

const cvaT2 = atomicTest({
  disabled: true,
  active: true,
})
```
Usage
```html
<span>Atomic Recipe</span>
<span className={cvaT1}>T1</span>
<span className={cvaT2}>T2</span>
```
image
image

---

With config recipes, all classes are used so the styles applied are determined by the order that they are generated in the CSS.

Definition
```ts
const test = defineRecipe({
  className: "test",
  variants: {
    active: {
      true: {
        bg: "red",
      },
    },
    disabled: {
      true: {
        bg: "blue",
      },
    },
  },
})
```

Usage
```ts
const recipeT1 = test({
  active: true,
  disabled: true,
})

const recipeT2 = test({
  disabled: true,
  active: true,
})
```
```html
<span>Config Recipe</span>
<span className={recipeT1}>T1</span>
<span className={recipeT2}>T2</span>
```

image
image

If you want for the `active` style to always apply, you have to remove the `disabled: true` variant. This is because `cvaT1` first determined the order the CSS is generated, if you always want to apply the `active` variant, then you have to be consistent with order you specify the used variants.

```ts
const recipeT1 = test({ // matched the order
  disabled: true,
  active: true,
})

const recipeT2 = test({
  disabled: true,
  active: true,
})
```

---

With this PR, you can specify `variantsPrecedence` in the definition.

```ts
const test = defineRecipe({
  className: "test",
  variants: {
    active: {
      true: {
        bg: "red",
      },
    },
    disabled: {
      true: {
        bg: "blue",
      },
    },
  },
  variantsPrecedence: ['disabled', 'active'],
})
```

It makes sure the active is generated last in the CSS resulting in `active` variant is always applied regardless of the order it is specified in the usage of the config recipe.

No


segunadebayo:
Hey, @cs-clarence. I appreciate this PR and the work done to understand the repo.

We had this idea earlier but decided not to add it to Panda since we already have the `compoundVariant` API. Here's what the recipe's refactor would look like.

```tsx
const testRecipe = defineRecipe({
  className: 'test',
  variants: {
    disabled: {
      true: {
        bg: 'red'
      }
    },
    active: {
      true: {}
    }
  },
  compoundVariants: [{ active: true, css: { bg: 'blue' } }]
})
```

Please use this approach instead. We don't want to introduce a new API.

Again, I sincerely appreciate this effort.

cs-clarence:
@segunadebayo understood, I am aware of the `compoundVariants` but decided to not heavily use it because disables the [responsive and conditional variants](https://panda-css.com/docs/concepts/recipes#responsive-and-conditional-variants) feature of config recipes which is very useful for me. If you could reconsider this PR or if you have plans to remove that limitation, I'd be happy to help.

PR #2919, opened 2024-10-13
fix(postcss): race condition on builder instance for simultaneous plugin invocations

doubleaxe:
Closes [#2916](https://github.com/chakra-ui/panda/discussions/2916)

## 📝 Description

Fixes postcss race condition because builder instance is singelton, and postcss plugin is async.

Reported in bug report.

## 🚀 New behavior

Added `builderGuard` which allow only one plugin invocation to be run at a time, other invocations will wait. Tested on my instance.

### DIFF
--- a/packages/postcss/src/index.ts
+++ b/packages/postcss/src/index.ts
@@ -1,7 +1,7 @@
 import { Builder, setLogStream } from '@pandacss/node'
 import { createRequire } from 'module'
 import path from 'path'
-import type { PluginCreator } from 'postcss'
+import type { PluginCreator, TransformCallback } from 'postcss'
 
 const customRequire = createRequire(__dirname)
 
@@ -20,7 +20,9 @@ export const loadConfig = () => interopDefault(customRequire('@pandacss/postcss'
 
 let stream: ReturnType<typeof setLogStream> | undefined
 
-const builder = new Builder()
+// export for unit test
+export const builder = new Builder()
+let builderGuard: Promise<void> | undefined
 
 export const pandacss: PluginCreator<PluginOptions> = (options = {}) => {
   const { configPath, cwd, logfile, allow } = options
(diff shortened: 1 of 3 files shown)


packages/postcss/src/index.ts
```diff
@@ -21,47 +21,61 @@ export const loadConfig = () => interopDefault(customRequire('@pandacss/postcss'
...
+          if (!builderGuard) break
+          await builderGuard
+        }
+        const _builderGuard = postcssProcess(...args)
+        if (!_builderGuard) return
+
+        builderGuard = _builderGuard.finally(() => {
+          builderGuard = undefined
         })
+        return builderGuard
```
astahmer, line 78: what is happening here ? I'm not familiar with this `for (;;)` trick

also, would it be possible to add a test somewhere in there https://github.com/chakra-ui/panda/blob/main/packages/postcss/__tests__/postcss.test.ts

xcarpentier:
Hi @doubleaxe,

Do you think that PR will fix other issue "I don't see any styles" or it's not related at all ?

https://panda-css.com/docs/installation/nextjs#i-dont-see-any-styles

Thanks

doubleaxe:
> Do you think that PR will fix other issue "I don't see any styles" or it's not related at all ?
> 
> https://panda-css.com/docs/installation/nextjs#i-dont-see-any-styles

Hello @xcarpentier.

I'm not familiar with nextjs, so I cannot tell for sure. The issue says about styles caching, maybe it is unrelated to this case.

packages/postcss/src/index.ts
doubleaxe: Well, this is heavily based on javascript promises, async, and single threaded nature. Trying to explain:

- When postcss executes method first time, `builderGuard` is undefined, so `if (!builderGuard) break` will be executed, then processing starts `const _builderGuard = postcssProcess(...args)` (note, no await, so `_builderGuard` is pending promise). `if (!_builderGuard) return` is just to make typescrupt happy. We assign promise to `builderGuard` global variable and add `_builderGuard.finally` to resets this global when promise is settled. Then we return this promise to postcss, and it starts waiting this processing to be finished.
- If builder is still processing, and another call arrives - `builderGuard` will be pending promise, so `await builderGuard` is executed and it waits for processing to be finished. Multiple calls may wait for this promise simultaneously.
- If processing is finished, `builderGuard` will be settled and global variable will be `undefined`, and one of the `await builderGuard` will continue. Because `builderGuard` is now undefined (in `finally`) cycle will break and another processing will start. All other simultaneous tasks will still wait, because of this `for (;;)` loop.

I added simple test, and it works, but it requires exposing builder instance. `export const builder`

astahmer: got it, so basically you're waiting from a global promise

could this work ? it seems a bit simpler
```ts
 async function (...args) {
      if (builderGuard) await builderGuard.catch(() => {});
      builderGuard = postcssProcess(...args);
      return builderGuard;
    },
```

doubleaxe: This variant will fix concurrent shared context initialization bug.

I think it will work correctly if there is only one concurrent invocation waiting. I.e. one waiting and one processing. If there are two waiting (in my project there were 3 concurrent invocations) - they will wait for first task to complete, and then they both will be executed concurrently.

It can work assuming there will be no other concurrent invocation bugs in builder, but to be on a safe side I'd rewrite it to something like:

```javascript
      function (...args) {
        builderGuard = Promise.resolve(builderGuard)
          .catch(() => {})
          .then(() => postcssProcess(...args))
        return builderGuard
      },
```

There are no await between reading and setting `builderGuard` so it will always point to last executed process.

doubleaxe: PR updated

astahmer (approved):

RobertoNegro:
I was having a similar problem in my project (Vike + SSR + Hono + Cloudflare Pages) as well. When I started the development server, the CSS with Panda's layers was almost empty (only the reset, base, and tokens layers were present), resulting in an unstyled page. I found that changing some settings in the Panda configuration (e.g., setting `minify` to `false` and then reverting the change) would solve the problem, so I suspected there was an issue during the initial HMR call. However, I had to do this every time I restarted the dev server. With your solution posted in #2916, I was able to start the development server with the correct style generation.

PR #2939, opened 2024-10-22
feat: update conditions

FilipePfluck:
## 📝 Description

Adds `aria-invalid` to the _valid and _invalid condition

Also adds _top, _right, _bottom, _left conditions. Those are quite handy when using ark-ui, so I figure I'd include them. Might remove them if mantainers judge it unecessary.

Also updated the docs which was outdated

No


packages/preset-base/src/conditions.ts
```diff
@@ -57,8 +57,8 @@ export const conditions = {
 
   indeterminate: '&:is(:indeterminate, [data-indeterminate], [aria-checked=mixed], [data-state="indeterminate"])',
   required: '&:is(:required, [data-required], [aria-required=true])',
-  valid: '&:is(:valid, [data-valid])',
-  invalid: '&:is(:invalid, [data-invalid])',
+  valid: '&:is(:valid, [data-valid], [aria-invalid="false"])',
+  invalid: '&:is(:invalid, [data-invalid], [aria-invalid="true"])',
```
segunadebayo, line 61: Let's add only these. The placement conditions can be added in your userland code

segunadebayo:
I discussed this with the team, and we decided not to add this PR for now.

Please update your panda config for now. We'll consider it in the future.

PR #3177, opened 2025-03-13
fix: generated `CssVarKeys` type cannot provide autocomplete correctly

hyp3rflow:
Closes # 

## 📝 Description

Fix issue where generated `CssVarKeys` type cannot provide autocomplete correctly

It provides suggestion as if it is just `string` type.

## 🚀 New behavior

Correctly provides declared css variable keys along with `string` type.

## 📝 Additional Information

You can easily see the difference of this PR in ts playground: https://www.typescriptlang.org/play/?#code/C4TwDgpgBGDCDO8BqBDATgORQW2gXigHIBrCEAWhUKgB8jSKAjaugCnmDQEsA7AcygAyKAG8AvgEoAUKEgwEydAGkyUAgANy5ACQi4iVJhwQx6qQHpzUawD0A-FJnhoPBYay419MpRbemhE5yrgbKqhpauiGKRrimtFCaOiIc3PzxwuIWVrZ2QA

### DIFF
--- a/packages/studio/styled-system/types/style-props.d.ts
+++ b/packages/studio/styled-system/types/style-props.d.ts
@@ -8,8 +8,8 @@ type AnyString = (string & {})
 type CssVars = `var(--${string})`
 type CssVarValue = ConditionalValue<Token | AnyString | (number & {})>
 
-type CssVarName =  | AnyString
-type CssVarKeys = `--${CssVarName}`
+type CssVarName = 
+type CssVarKeys = `--${CssVarName}` | `--${string}` & {}
 
 export type CssVarProperties = {
   [key in CssVarKeys]?: CssVarValue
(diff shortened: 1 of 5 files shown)


packages/studio/styled-system/types/style-props.d.ts
```diff
@@ -8,8 +8,8 @@ type AnyString = (string & {})
 type CssVars = `var(--${string})`
 type CssVarValue = ConditionalValue<Token | AnyString | (number & {})>
 
-type CssVarName =  | AnyString
-type CssVarKeys = `--${CssVarName}`
+type CssVarName =
```
hyp3rflow, line 11: How can I prevent this empty type? I tried to fix this by editting `style-props.ts`, but it didn't work.

segunadebayo:
Thanks for this @hyp3rflow. I've cherry-picked some changes into the main branch.

We're release an update shortly

PR #3182, opened 2025-03-17
fix: normalize path separators

juanrgm:
## 📝 Description

The default `cwd` option is `process.cwd()` and it uses Windows path separators, but there is code that uses the unix path separator for string matching.

Some import declarations are not detected if you use tsconfig path aliases.

## 🚀 New behavior

The import declarations are detected and the result of parser is not empty.

No.

## 📝 Additional Information

A temporary workaround could be:

```ts
// panda.config.ts
export default defineConfig({
  cwd: process.cwd().toString().replaceAll("\\", "/")
});
```

### DIFF
--- a/.changeset/plenty-kings-collect.md
+++ b/.changeset/plenty-kings-collect.md
@@ -1,0 +1,5 @@
+---
+'@pandacss/core': patch
+---
+
+Fix import detection in Windows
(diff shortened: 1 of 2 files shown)


juanrgm:
Any news?

PR #3279, opened 2025-06-12
refactor(system): improve property priority

### LINKED ISSUE #3259: Priority of styles inconsistent based on order of generated css
### Description

We're making components that accept `css` props so we can merge them with our component's recipes, [as suggested in the documentation](https://panda-css.com/docs/concepts/merging-styles#merging-cva--css-styles).

Our problem is that if our recipe includes a style like `padding: '16px'`, and someone tries to use the `css` prop to set a style like `paddingX: '2px'`, the priority will be determined by the order that the styles are picked up by Panda's parser. I'm worried that this will cause hard-to-find bugs.

### Link to Reproduction

https://play.panda-css.com/lzOJ2mjnGv

### Steps to reproduce

1. Comment out line 7
2. Notice that unrelated styles change

### JS Framework

_No response_

### Panda CSS Version

0.53.6

### Browser

Safari 18.5

### Operating System

- [x] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Maybe the output CSS should always be ordered in a more deterministic way? Ideally, more specific styles like `paddingX` should always take higher priority than `padding` - similar to [Tailwind](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier#how-classes-are-sorted)

segunadebayo:
Closes #3259

## 📝 Description

Improve algorithm for deterministic property order.

- Longhand (`padding`, `margin`, `inset`)
- Shorthand of longhands (`padding-inline`, `margin-inline`)
- Shorthand of shorthands (`padding-inline-start`, `margin-inline-start`)

```tsx
css({
  p: '4',
  pr: '2',
  px: '10',
})
```

Will result in the following css regardless of the order of the properties:

```css
.p-4 {
  padding: 4px;
}

.px-10 {
  padding-left: 10px;
  padding-right: 10px;
}

.pr-2 {
  padding-right: 2px;
}
```

Non-deterministic (in some cases)

## 🚀 New behavior

Fairly more deterministic

Potentially, depending on how the user has designed their styles.


segunadebayo:
**Before**
CleanShot 2025-06-12 at 14 30 05@2x

**After**
CleanShot 2025-06-12 at 14 29 11@2x

PR #3290, opened 2025-06-20
Update template-literals.md

s-h-a-d-o-w:
## 📝 Description

(This is related to the discussion https://github.com/chakra-ui/panda/discussions/3277 )

Currently, these docs mention that template literals are an alternative to object literals but there are actually no docs for setting up object literals except for the styled components migration docs.

In my opinion, styled components and emotion have supported object literals for so many years that the two approaches go hand in hand. And so I think it makes sense to point out how to set up both, independently from the migration section. Especially since I would argue that this process is more straightforward if one already has a basic panda setup.

(I felt it would be best to just show a WIP like this. I understand that if this was merged as is, it might not look pretty in the sidebar. Or maybe you think that this should be structured/done differently.)

### DIFF
--- a/website/pages/docs/concepts/template-literals.md
+++ b/website/pages/docs/concepts/template-literals.md
@@ -1,11 +1,11 @@
 ---
-title: Template Literals
-description: Panda allows you to write styles using template literals.
+title: Template/Object Literals
+description: Panda allows you to write styles using template or object literals.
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks for this PR. I'll reword it in the main branch

PR #3291, opened 2025-06-22
fix: change slot recipe raw function typings

firzanarmani:
## 📝 Description

`SlotRecipeRuntimeFn`'s `raw()` function uses a return type that seems unnecessary to be different from the props' type; especially since the respective js generated (as seen in generator artifacts) is simply just `raw: (props) => props`. Correcting this would make working with slot recipe types easier as you can map the generic types to the actual slot recipe types that are being generated.

The generic SlotRecipeRuntimeFn's `raw()` return type is `Record<S, SystemStyleObject>` which is not a proper match to `RecipeSelection<T>`

## 🚀 New behavior

Return type uses the `RecipeSelection<T>`.

No


segunadebayo:
This is incorrect. `sva.raw(...)` actually returns the css object for each slot.

PR #3405, opened 2025-10-31
docs: fix nextjs panda.config example

mateusvelleda:
Closes # 

## 📝 Description

NextJS docs bad formatting on panda.config.ts

Instalação Next js Panda CSS

## 🚀 New behavior

Well formatted

## 📝 Additional Information

I think it's an issue with VS Code Format on Save function once the prettier ignore command has no effect on it
The only way I found was to make the changes then save using Cmd+Shift+P > File: Save without Formatting
Even after saved, if I press Cmd+S it gets bad formatting again


segunadebayo:
Thanks for this PR. Looked into the docs and fixed even more related issues

PR #3479, opened 2026-02-14
feat: support for dynamic conditions

anubra266:
Add support for dynamic conditions


segunadebayo:
After thoroughly reviewing this, we need to take a different approach to the solution so it works in v2's Rust Engine as well.

I'll port this PR into v2 this weekend. Thanks

PR #3511, opened 2026-04-14
fix: extract compiled jsx styles

### LINKED ISSUE #3509: `css` prop inline object literals not extracted from compiled (dist) files
### Description

Panda fails to extract styles when a `css` prop is passed an inline object literal in compiled (dist) files. The same pattern is extracted correctly from source files.
This was discovered while using Panda to tree-shake CSS across both local source files and compiled library dist files (e.g. component libraries from a design system).

When JSX is compiled, `<Box css={{ color: 'RebeccaPurple' }} />` becomes a `jsx()` function call. Panda's static analysis appears to handle the JSX form correctly, but does not recognize the inline object literal in the compiled form.

| Pattern | Source | Dist |
|---|---|---|
| Inline object literal `css={{ color: '...' }}` | ✅ Extracted | ❌ Not extracted |
| Pre-defined variable `css={styles}` | ✅ Extracted | ✅ Extracted |
| Spread of pre-defined variable `css={{ ...spreadStyles }}` | ✅ Extracted | ✅ Extracted |

### Link to Reproduction

https://stackblitz.com/edit/vitejs-vite-4pyiuwzk?file=src%2FApp.tsx&terminal=dev

### Steps to reproduce

Source file (works):
```jsx
import type { ReactNode } from 'react';

import { css, cx } from '@/styled-system/css';
import type { SystemStyleObject } from '@/styled-system/types';

interface BoxProps {
  children?: ReactNode;
  css?: SystemStyleObject;
}

const Box = (props: BoxProps) => {
  const { css: cssProp, children } = props;
  return <div className={cx(css(cssProp))}>{children}</div>;
};

const styles = css.raw({
  color: 'white',
  backgroundColor: 'AntiqueWhite',
});

const spreadStyles = css.raw({
  color: 'white',
  backgroundColor: 'SteelBlue',
});

export const CustomComponent = () => {
  return (
    <>
      {/* Panda CAN extract from source, but CANNOT from dist */}
      <Box css={{ color: 'black', backgroundColor: 'RebeccaPurple' }}>Box</Box>

      {/* Panda CAN extract from source and dist */}
      <Box css={styles}>Box</Box>
      <Box css={{ ...spreadStyles }}>Box</Box>
    </>
  );
};
```

Compiled dist file (broken):
```js
import { Fragment, jsx, jsxs } from "react/jsx-runtime";
import { css, cx } from "@/styled-system/css";

const Box = (props)=>{
    const { css: cssProp, children } = props;
    return /*#__PURE__*/ jsx("div", {
        className: cx(css(cssProp)),
        children: children
    });
};
const styles = css.raw({
    color: 'white',
    backgroundColor: 'AntiqueWhite'
});
const spreadStyles = css.raw({
    color: 'white',
    backgroundColor: 'SteelBlue'
});
const CustomComponent = ()=>{
    return /*#__PURE__*/ jsxs(Fragment, {
        children: [
            /*#__PURE__*/ jsx(Box, {
                css: {
                    color: 'black',
                    backgroundColor: 'RebeccaPurple'
                },
                children: "Box"
            }),
            /*#__PURE__*/ jsx(Box, {
                css: styles,
                children: "Box"
            }),
            /*#__PURE__*/ jsx(Box, {
                css: {
                    ...spreadStyles
                },
                children: "Box"
            })
        ]
    });
};

export { CustomComponent };
```

Panda should extract styles from inline object literals passed to the `css` prop regardless of whether the file is a source `.tsx` file or a compiled `.js` dist file.

### JS Framework

React (TS)

### Panda CSS Version

1.8.2

### Browser

_No response_

### Operating System

- [x] macOS
- [ ] Windows
- [x] Linux

### Additional Information

_No response_

kris-ellery:
Closes #3509

## 📝 Description

Panda's static extractor now recognizes compiled JSX factory calls (`jsx`, `_jsx`, `jsxs`, `_jsxs`, `createElement`) and extracts inline style objects from them.

When bundlers compile JSX, `<Box css={{ color: "red.900" }} />` becomes `jsx(Box, { css: { color: "red.900" } })`. The extractor only handled JSX element syntax, so inline style objects in compiled output were silently dropped. This affects any library that ships compiled dist files and uses inline object literals with the `css` prop.

The two other patterns already worked:
- `css={styles}` (variable via `css.raw()`) — extracted at the definition site
- `css={{ ...styles }}` (spread) — same reason

## 🚀 New behavior

A new detection block inside the existing `components` handler recognizes compiled JSX call expressions and extracts the component name from the first argument and style props from the second argument. It produces `kind: 'component'` results, so the existing parser handles them with zero changes.

```js
// Now extracted correctly
jsx(Box, { css: { color: "red.900", backgroundColor: "red.200" } })
jsxs(Box, { css: { padding: "4" }, children: ["hello"] })
React.createElement(Box, { css: { padding: "4" } })
```

Non-style props (`children`, `onClick`, `key`, etc.) are filtered during extraction via `matchProp` so they are never evaluated.

No

## 📝 Additional Information

- No parser changes needed — existing component result handler picks up the new extraction results automatically
- No CSS output changes — core `atomic-rule` snapshot tests pass with no diffs
- 7 extractor-level tests + 7 parser integration tests added, including a realistic compiled dist scenario matching the reproduction in #3509

### DIFF
--- a/.changeset/extract-compiled-jsx-styles.md
+++ b/.changeset/extract-compiled-jsx-styles.md
@@ -1,0 +1,10 @@
+---
+'@pandacss/extractor': patch
+---
+
+Fix static extraction of inline style objects from compiled JSX calls.
+
+Panda's extractor now recognizes compiled JSX factory calls (`jsx`, `_jsx`, `jsxs`, `_jsxs`, `createElement`) and
+extracts inline style props from the second argument. Previously, styles passed as inline object literals in compiled
+output (e.g. `jsx(Box, { css: { color: "red.900" } })`) were silently dropped because the extractor only handled
+JSX element syntax, not the equivalent call expression form produced by bundlers.
--- a/packages/extractor/__tests__/extract.test.ts
+++ b/packages/extractor/__tests__/extract.test.ts
@@ -6411,3 +6411,126 @@ it.skip('extracts slots when spread', () => {
     }
   `)
 })
+
+const compiledJsxConfig: Record<string, string[]> = {
+  Box: ['css', 'color', 'bg'],
+  'styled.div': ['bg', 'color'],
+}
+
+const compiledJsxMatcher: ComponentMatchers = {
+  matchTag: ({ tagName }) => Boolean(compiledJsxConfig[tagName]),
+  matchProp: ({ tagName, propName }) => compiledJsxConfig[tagName]?.includes(propName) ?? false,
+}
+
+it('compiled jsx - extracts from realistic compiled dist output', () => {
+  const result = extractFromCode(
+    `
+    import { Fragment, jsx, jsxs } from 'react/jsx-runtime';
+
+    const App = () => {
+      return jsxs(Fragment, {
+        children: [
+          jsx(Box, {
+            css: {
+              color: 'red.900',
+              backgroundColor: 'red.200',
+            },
+            children: 'Box',
+          }),
+        ],
+      });
+    };
+    `,
+    { components: compiledJsxMatcher },
+  )
+
+  expect(Object.keys(result)).toContain('Box')
+  expect(result.Box[0].raw).toHaveProperty('css')
+  expect(result.Box[0].raw).not.toHaveProperty('children')
+})
+
+it('compiled jsx - extracts inline object from _jsx(Component, { prop: value })', () => {
+  const result = extractFromCode(
+    `
+    function render() {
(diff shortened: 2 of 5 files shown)


segunadebayo:
Closing based on https://github.com/chakra-ui/panda/issues/3509#issuecomment-4318609799

PR #3523, opened 2026-04-28
feat(generator): add createRecipeContext and createSlotRecipeContext for react jsx

Adebesin-Cell:
> **Note:** prep work for chakra static + panda. this PR adds the jsx helpers that the chakra static integration will sit on top of. landing here so chakra can adopt them without forking panda.

emits two new helpers into styled-system/jsx for react projects. usable standalone by anyone building headless components on panda.

Closes #

## 📝 Description

adds `createRecipeContext` and `createSlotRecipeContext` to the generated `styled-system/jsx` entry, alongside the existing `createStyleContext`. both are emitted only when the jsx framework is `react`. they accept a recipe directly, `{ key: 'name' }` to look up a config recipe from `../recipes`, or `{ recipe }`.

`styled-system/jsx` exposes `createStyleContext` for slot recipes only. there is no equivalent helper for non-slot recipes, no `PropsProvider` for cascading default variant props, and no built-in way to bind a component to a config recipe by name without importing the recipe yourself.

## 🚀 New behavior

`createRecipeContext(recipe)` returns `{ withContext, PropsProvider, usePropsContext }`. wraps a single (non-slot) recipe so a styled component picks up the recipe's variant props as jsx props, supports `unstyled`, and lets a parent inject default variant props through `PropsProvider` without prop drilling.

`createSlotRecipeContext(slotRecipe)` returns `{ withRootProvider, withProvider, withContext }`. runs a slot recipe once at the root, stores the resolved slot styles in a react context, and lets each slot consume them without re-running the recipe.

```tsx
import { createSlotRecipeContext } from 'styled-system/jsx'
import { menu } from 'styled-system/recipes'

const ctx = createSlotRecipeContext(menu)
export const MenuRoot = ctx.withRootProvider(Ark.Menu.Root)
export const MenuItem = ctx.withContext(Ark.Menu.Item, 'item')
```

No. additive only. existing `createStyleContext` is untouched.

## 📝 Additional Information

- react only for now. solid / vue / preact can mirror this later by adding entries to `createRecipeContextMap` / `createSlotRecipeContextMap` in `packages/generator/src/artifacts/jsx.ts`, the same way `create-style-context` does it.
- recipes-by-key lookup is gated on `!ctx.recipes.isEmpty()` so projects with no config recipes still emit valid output.
- two new `ArtifactId`s (`jsx-create-recipe-context`, `jsx-create-slot-recipe-context`) with matching `config-deps` entries.
- `pnpm test packages/generator` — snapshot updated to include the new files.
- `pnpm test packages/core` — 234/234 pass, css output unchanged.


segunadebayo:
Thanks for this PR. I'll add it to the Rust port directly.

PR #3531, opened 2026-05-04
feat: add grouped styles mode via `cssMode` config option

gajus:
## Motivation

Production Panda apps routinely produce class sequences like:

```html
class="gGTyfw cjSpVR VxsAK kYufQm iKPHnu cPHhKG bYPztT jZzPrs hYCnIA dDifmo
       hDGJaH dTBYkx bjOqtc bZRhvx fPSBzf bYPznK diIxfU jTWvec jLRTbm
       kFYoVQ fNDbsJ fNDbsm"
```

Atomic CSS is a good default, but at scale it has measurable costs that `hash: true` doesn't address (it shortens class names, not their count). This PR adds an opt-in flag to trade some CSS duplication for smaller, more inspectable HTML.

## Evidence

**Style recalc scales with declaration count.** Per [web.dev](https://web.dev/articles/reduce-the-scope-and-complexity-of-style-calculations), "the worst case cost of calculating the computed element's style is the number of elements multiplied by the selector count." A [Dec 2025 benchmark](https://dev.to/yoskutik/is-tailwind-actually-slow-533d) compared atomic vs. grouped CSS on identical visual output (5K components, 20K DOM nodes, same CSS file size). Atomic produced 35K declarations vs. 10K grouped:

| Browser | Atomic | Grouped | Delta |
|---------|--------|---------|-------|
| Chrome  | 169 ms | 121 ms  | +28%  |
| Safari  | 406 ms | 272 ms  | +33%  |
| Firefox | 114 ms | 82 ms   | +25%  |

Padding class names to equalize HTML size left the gap intact – declaration count was the driver.

**HTML parse cost scales with bytes, even after compression** (the parser sees the decompressed string). The same benchmark generated 1.1 MB of HTML for atomic vs. 0.4 MB for grouped. [Peterbe's case study](https://www.peterbe.com/plog/how-much-html-is-too-much-webperf) measured 119KB vs. 31KB HTML on the same page: parse + layout dropped from 523ms to 126ms under throttled conditions.

**Dev tools UX.** Acknowledged in Panda's own docs: long atomic classNames make DOM inspection painful. `hash` and `groupedStyles` solve different parts of this.

This matters most for SSR/SSG, where HTML can rarely be aggressively cached and lands on every initial render.

## Tradeoff

`cssMode: 'grouped'` increases CSS bundle size – shared properties across `css()` calls no longer dedupe to single atomic classes. The right call depends on whether the bottleneck is HTML payload + recalc time or CSS bundle size. That's why it's opt-in.

## What this PR does NOT change

Atomic remains the default. No breaking changes. Cascade layer ordering, `hash`, `cva`, recipes, and patterns are unaffected.

## Implementation

Minimal changes localized to the extractor, behind a single config flag.


astahmer:
fwiw I thought about adding it at some point but figured there were too many caveats (see the PR's description) for it to be worth adding into the core package

https://github.com/chakra-ui/panda/pull/1492

gajus:
> fwiw I thought about adding it at some point but figured there were too many caveats (see the PR's description) for it to be worth adding into the core package
> 
> #1492

Thanks for sharing.

The main issues called out were ternaries and `css.raw` merging, where the parser splits a single `css()` call into multiple data entries that don't match what the runtime evaluates. These are handled now:

  - Ternaries: The build step reconstructs the combinations the runtime would see by separating base properties from branch alternatives and generating the cartesian product. `css({ fontSize: "xl", color: active ? "red" : "blue" })` produces groups for both `{ fontSize: "xl", color:  "red" }` and `{ fontSize: "xl", color: "blue" }`.
  - `css.raw` merging: Non-overlapping entries are merged into a single group at build time, matching what the runtime produces after `Object.assign`.

Everything is behind cssMode: "grouped" so it's opt-in with no impact on existing behavior.

If I missed something, would appreciate if you share a breaking test case.

gajus:
@segunadebayo any comments?

gajus:
Struggling to get some feedback here. Maybe @anubra266 ?

segunadebayo:
Thanks for the PR @gajus. Will have a look over the weekend and let you know

gajus:
Thank you.

For what it is worth, I've used Claude to brainstorm edge case/test pairs, and could not identify any undesirable behaviors.

gajus:
@segunadebayo would really like to see this land in the next version. Let me know if any further input is needed from my end.

gajus:
@segunadebayo any update here?

gajus:
@segunadebayo Are there any other maintainers I can tag here to accelerate the resolution?

jfreehill:
Hello, I'm also curious to know about any movement/interest on this. The initial page-render impact potential for repeated elements with many one-off classes seems realistic. 

I wonder if it would be valuable to simplify a grouping functionality by grouping rules that are already flagged as one-offs via the escape-hatch syntax (enforced via `strictTokens`)? In my experience so far, the most verbose class lists are those with one-offs or very specific rule sets.

segunadebayo:
@gajus This PR is definitely valuable. I'm more inclined to add this to v2 rather than v1 so we can better shape the config.

Trying to see if we can avoid a new config property and potentially combine with any existing one.

PR #3541, opened 2026-05-09
fix(postcss): strip @layer order declaration when polyfill is enabled

### LINKED ISSUE #3540: PostCSS: @layer order declaration is not stripped from CSS output when polyfill: true
### Description

When I set polyfill: true in `panda.config.ts` and process my entry CSS file through `@pandacss/dev/postcss`, I expected the final CSS output to contain no @layer syntax at all — since the whole point of the polyfill is to make the output safe for browsers that don't support cascade layers.

index.css: 
``` index.css
@layer reset, base, tokens, recipes, utilities;
```

  But what happened instead is that the original @layer reset, base, tokens, recipes, utilities; order declaration from my source CSS is preserved in the final output. 

output: 
``` index.css
@layer reset, base, tokens, recipes, utilities;

:root {
  --made-with-panda: "🐼";
}

*, :before, :after, ::backdrop {
  --blur: /*-*/ /*-*/;
  --brightness: /*-*/ /*-*/;
...
```

### Link to Reproduction

-

### Steps to reproduce

1. follow the https://panda-css.com/docs/installation/rsbuild#start-a-new-project to init。
2. add `polyfill: true` in `panda.config.ts` and run build.

### JS Framework

React & TS

### Panda CSS Version

1.11.1

### Browser

-

### Operating System

- [x] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

Because I’m the provider of a component library, other apps may use my components and compile their CSS with `@csstools/postcss-cascade-layers` or `postcss-preset-env`. In that case, the reserved `@layer reset, base, tokens, recipes, utilities;` causes my CSS to be expanded with a large number of unexpected `not(#\#)` selectors, which makes my styles much higher priority than intended.

tamish-max:
Fixes #3540

## Problem

When polyfill: true is set in panda config, the PostCSS plugin correctly applies @csstools/postcss-cascade-layers to the **generated** CSS (in sheet.toCss()). However, the user's entry CSS still contains the @layer reset, base, tokens, recipes, utilities; order declaration — it is never removed from the PostCSS root.

This orphaned declaration causes downstream polyfill tools (like postcss-preset-env) to produce bloated :not(#\#) specificity selectors on all styles, because they try to rewrite the order declaration but find no matching @layer blocks.

## Fix

Strip all @layer at-rules from the PostCSS root before appending the polyfilled CSS, when polyfill: true is enabled. This is safe because:
1. The generated CSS has already been polyfilled (no @layer rules remain)
2. The CSS is emitted in correct cascade order (reset → utilities)
3. Only the user's entry CSS is in the root at this point

### DIFF
--- a/packages/node/src/builder.ts
+++ b/packages/node/src/builder.ts
@@ -196,6 +196,17 @@ export class Builder {
     ctx.appendBaselineCss(sheet)
     const css = ctx.getCss(sheet)
 
+    // When polyfill is enabled, the generated CSS has already been processed
+    // by @csstools/postcss-cascade-layers (in sheet.toCss()). The user's
+    // entry CSS still contains the @layer order declaration which must be
+    // removed to prevent downstream polyfill tools from producing bloated
+    // :not(#\#) specificity selectors.
+    if (ctx.config.polyfill) {
+      root.walkAtRules('layer', (rule) => {
+        rule.remove()
+      })
+    }
+
     root.append(css)
   }
 
--- a/packages/postcss/__tests__/postcss.test.ts
+++ b/packages/postcss/__tests__/postcss.test.ts
@@ -49,6 +49,18 @@ describe('PostCSS plugin', () => {
     expect(result.css.length).toBeGreaterThan(2)
   })
 
+  test('strip layer declarations when polyfill is enabled', async () => {
+    const input = '@layer reset, base, tokens, recipes, utilities;'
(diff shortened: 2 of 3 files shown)


segunadebayo:
Thanks for the PR. This already landed on main in a tighter form: we only strip the bare `@layer reset, base, tokens, recipes, utilities;` order declaration, not every `@layer` in the user's entry file.

Closing as already shipped.

PR #3543, opened 2026-05-15
chore: add engines field to package.json

gameroman:
This makes it easier for consumers to know what to expect as far as compatibility goes.

Both human consumers and static analysis tools.


Adebesin-Cell (commented):
Could you add a description to the PR about why?

Adebesin-Cell:
Thanks for updating the description! One more thing: the `>=20.12.0` floor itself, is there a specific reason for that version (a feature you're using, alignment with our CI, an LTS cutoff)? It would be good to call it out so the number isn't picked out of thin air.

And a changeset would help here too, since this changes a published manifest.

gameroman:
>  the `>=20.12.0` floor itself, is there a specific reason for that version (a feature you're using, alignment with our CI, an LTS cutoff)?

I based it on LTS and the JS features and dependencies you use where this version seems to be the one that is supported

> And a changeset would help here too, since this changes a published manifest.

Added changeset

Adebesin-Cell:
I think we should pin to version 20.x that would make it less strict

packages/cli/package.json
```diff
@@ -74,5 +74,8 @@
     "@types/update-notifier": "6.0.8",
     "kleur": "4.1.5",
     "update-notifier": "7.3.1"
+  },
+  "engines": {
+    "node": ">=20.12.0"
```
gameroman:
```suggestion
    "node": ">=20"
```

Adebesin-Cell (approved):

PR #3558, opened 2026-06-03
feat(v2): esm-only packages

Adebesin-Cell:
> **Note:** v2 only. all packages go ESM-only — `"type": "module"`, single `.js` build, no `require` condition. CJS consumers keep working through node's `require(esm)`.

## 📝 Description

removes the dual cjs+esm build from every published package. 30 manifests flip to `"type": "module"` with `engines.node >=22.12.0`, exports maps collapse to `source` / `types` / `default`, `main` and `module` fields are gone. tsup builds `--format=esm` only, shims dropped everywhere except the compiler-wasm node entry (its relative dynamic import still needs the `__dirname` shim).

every package ships `dist/index.js` (cjs) + `dist/index.mjs` (esm) plus paired `.d.ts`/`.d.mts`. bins and the `dev/postcss` re-export are cjs. `--shims` papers over `__dirname`/`require` in esm output.

## 🚀 New behavior

- single esm build per package: `dist/index.js` + `dist/index.d.ts`
- `require('@pandacss/postcss')` / `require('@pandacss/dev/postcss')` return the plugin function directly via node's `module.exports` named export — the next.js webpack path (`postcss.config.cjs`) works without unwrapping
- `update-notifier` + `kleur` become regular dependencies of `@pandacss/dev` instead of being bundled (bundled cjs deps hit esbuild's throwing require stub under pure esm; they were also imported from shipped src while sitting in devDependencies)
- bins are esm, package scripts use `import.meta.dirname`

## 📊 numbers (upstream/v2 baseline vs this branch, same machine)

| | dual cjs+esm | esm-only | delta |
|---|---|---|---|
| total dist | 7,368 KB | 3,784 KB | **−49%** |
| dist files | 212 | 123 | −42% |
| `@pandacss/dev` dist | 944 KB | 96 KB | **−90%** |
| `@pandacss/core` dist | 340 KB | 168 KB | −51% |
| full `pnpm build` | ~45.8s | ~43.8s | ~same |

Yes. node `>= 22.12.0` required (the `require(esm)` support line; node 20 is EOL). cjs consumers don't need code changes — verified against the next-js-app and vite sandboxes and a `require()` sweep of all 29 packages from cjs (also proves no top-level await in any entry graph).

## 📝 Additional Information

good to know:

- exports condition order is `source` → `types` → `default`, not publint's types-first. deliberate: the workspace builds with `customConditions: ["source"]` and types-first shadows it, breaking sibling dts builds. consumers never enable `source`, so resolution is unchanged for them. publint may warn; accepted.
- two pre-existing baseline bugs fixed in passing: the compiler/compiler-wasm dts build race (scoped `tsconfig.build.json`, same pattern as node/config) and a stale `specifiers` option in cli_v2 after the `codegenImportExtensions` rename.
- prettier 3.2.5 can't parse `export { x as 'module.exports' }`, so the two interop files are prettierignored with a removal note. follow-up: bump prettier, then drop the ignore entries.
- `pnpm typecheck` is red on upstream/v2 already (33 condition-type errors from the block condition lowering work). unrelated to this change, untouched here.
- config bundle snapshots updated `.mjs` → `.js` (8 hunks, paths only). css output untouched — atomic-rule and the full core suite pass at baseline counts.
- follow-ups not in this PR: publint + attw in CI, website docs on the node floor, prettier bump.

## test plan

- [x] `pnpm build` ×2 green (deterministic after the race fix)
- [x] `pnpm test packages/core` 234/234 (css output identical), generator 71/71, parser 235/235, node 24/24, config 74/74, cli_v2 9/9
- [x] `require()` sweep: all 29 published packages load from cjs on node 22
- [x] publint clean on representative packages (compiler-wasm warns only about gitignored wasm artifacts)
- [x] sandboxes: next-js-app (webpack + `postcss.config.cjs`), vite, codegen (212 tests)


Adebesin-Cell:
Fixed in https://github.com/chakra-ui/panda/commit/b4157a5d2a107885dea9981042ea968ca13901a6

PR #3563, opened 2026-06-04
fix(mcp): point generated mcp config at the @pandacss/dev binary

Adebesin-Cell:
## what

`panda init-mcp` generated a server config that runs `npx panda mcp`. that command can't run: `panda` is an unrelated package on npm with no bin. the generated config now runs `npx -y --package @pandacss/dev panda mcp`.

## why

AI clients spawn the generated command to start the mcp server. with `npx panda mcp` they fail with `could not determine executable to run`, so the server never starts. came in as a user report. `bunx panda mcp` fails the same way.

## notes

- the same args work under `bunx`. npx ships with node, so pnpm/yarn users still get a runnable command without per-manager handling.
- this is the npx half of the report. the other half (config loading throwing `Please pass in filename to use require` under bun) is a separate `bundle-n-require`/`node-eval` issue and is not in this PR.

## test plan

- [ ] `pnpm test packages/mcp` (new `clients.test.ts`, 2 tests)
- [ ] `npx -y --package @pandacss/dev panda mcp` resolves the binary in a clean dir (only fails on the missing config, as expected)
- [ ] `bunx --package @pandacss/dev panda mcp` and `pnpm --package=@pandacss/dev dlx panda mcp` resolve too; `npx panda mcp` / `bunx panda mcp` reproduce the original failure

### DIFF
--- a/.changeset/mcp-npx-command.md
+++ b/.changeset/mcp-npx-command.md
@@ -2,6 +2,9 @@
 '@pandacss/mcp': patch
 ---
 
-Fix `panda init-mcp` generating a server config that AI clients can't run. The config used `npx panda mcp`, but `panda` is an unrelated package on npm with no executable, so clients failed with "could not determine executable to run" (the same happens with `bunx panda mcp`).
+Fix `panda init-mcp` generating an MCP server config that AI clients couldn't run, or that couldn't find your Panda
+config.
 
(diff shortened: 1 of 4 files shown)


Adebesin-Cell:
Closing would be superseded by https://github.com/chakra-ui/panda/pull/3677

PR #3595, opened 2026-06-18
fix(node): skip writing unchanged artifacts in OutputEngine

leegunwoo98:
## Problem

`OutputEngine.write` writes every generated artifact unconditionally:

```ts
return this.fs.writeFile(absPath, code)
```

`panda --watch` runs a full `generate()` on its initial build (before entering the watch loop), and codegen re-runs on config changes. When the output hasn't changed, these still **rewrite byte-identical files**, bumping their mtime. Build tools watch the *filesystem event*, not the content, so they react to these no-op writes:

- **Vite** fires a full page reload for the re-emitted `styled-system/*.mjs` (a reload "storm" right after `vite dev` boot) and an `(ssr) page reload` for the SSR importer of `styles.css`.
- **webpack** performs a redundant second compilation (see #3110).

## Fix

Skip the write when the artifact is byte-identical to what's already on disk, reusing the existing `Runtime['fs'].readFileSync` (utf8):

```ts
try {
  if (this.fs.readFileSync(absPath) === code) return
} catch {
  // file does not exist yet — fall through and write
}
```

`OutputEngine.write` is the single chokepoint for all artifact writes (codegen `.mjs`/`.d.ts`, `styles.css`, split CSS, config), so this covers them all. `ensureDirSync` still runs; first-write / `--clean` / changed-content all still write (a missing file throws → falls through to the write). Emitted output is unchanged — only redundant identical writes (and their mtime bumps) are skipped.

This is the standard "write-if-changed" pattern (TypeScript's internal `writeFileIfUpdated`, GraphQL Code Generator's unchanged-file skip, CMake `configure_file`, …) applied to Panda's artifact writer.

A changeset (`@pandacss/node: patch`) is included.

Refs #3110.

### DIFF
--- a/.changeset/skip-unchanged-artifacts.md
+++ b/.changeset/skip-unchanged-artifacts.md
@@ -1,0 +1,7 @@
+---
+'@pandacss/node': patch
+---
+
+Skip writing a generated artifact when its content is byte-identical to what is already on disk.
+
+`panda --watch` re-emits the full `styled-system` output on its initial build, and codegen re-runs on config changes. Rewriting unchanged files bumped their mtime and triggered needless dev-server reloads (e.g. a Vite full page reload or webpack recompile) for tools watching the output directory.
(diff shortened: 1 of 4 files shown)


leegunwoo98:
Superseded by #3597 — thanks @segunadebayo for the complete fix (routing `bundleStyles` through `writeCss` + tests). Closing in favor of that. 🙏

PR #3601, opened 2026-06-18
fix(compiler): name object-map utility classes by the author key, not a token alias

leegunwoo98:
## Problem

The native (Rust) engine and the runtime `css()` disagree on the class **name** for object-map utility values, so cssgen writes a rule the runtime never matches and the style silently does nothing.

```ts
css({ minHeight: '100vh' })
// element gets class:  min-h_100vh   (runtime)
// stylesheet contains: .min-h_screen  (cssgen)  ← no .min-h_100vh rule → min-height never applies
```

This affects every utility whose `values` is an object map that maps an alias to a literal also writable directly — in `@pandacss/preset-base` that's `width`/`minWidth`/`maxWidth` and `height`/`minHeight`/`maxHeight` (their value maps include `screen: '100vw'` / `screen: '100vh'`), plus the general case:

| Authored | Element class (runtime) | cssgen rule (before) | Match |
| --- | --- | --- | --- |
| `minHeight: '100vh'` | `min-h_100vh` | `.min-h_screen` | ❌ |
| `width: '100vw'` | `w_100vw` | `.w_screen` | ❌ |
| `maxWidth: '100vw'` | `max-w_100vw` | `.max-w_screen` | ❌ |
| `marginBottom: '0.5rem'`¹ | `mb_0.5rem` | `.mb_2` | ❌ |
| `marginBottom: '2'` (the key) | `mb_2` | `.mb_2` | ✅ |
| `width: '100%'` | `w_100%` | `.w_100%` | ✅ ² |

¹ with `marginBottom.values = { '2': '0.5rem' }` &nbsp;&nbsp; ² only "worked" by luck — `full` resolves to `var(--sizes-full)`, not the literal `100%`, so it didn't match the reverse-map scan.

## Root cause

The runtime hashes the class name as `propKey + withoutSpace(value)` **verbatim** — it has no token dictionary and never reverse-maps. The native engine, however, did two things that lost the author's literal:

1. The encoder's `normalize_property_value_cow` substituted the **resolved CSS** onto the atom (`'2'` → `'0.5rem'`, `'sm'` → `'4px'`).
2. To compensate, `class_name_value` then scanned `values` and reverse-mapped a resolved value **back** to its alias when naming the class.

Step 2 was lossy and order-dependent: it correctly restored `mb_2` for `marginBottom: '0.5rem'`, but for `minHeight: '100vh'` it reverse-mapped the *authored* literal `100vh` to the `screen` alias, producing `.min-h_screen` for an element labelled `min-h_100vh`.

## Fix

Keep the author's value-map **key** on the atom and resolve it to CSS only at emit time (`raw_property_value` / `default_style` already look the key up in `values` when writing declarations). The class name is then the literal the author wrote — matching the runtime, and matching **legacy Panda**, which named these by the literal too (verified against `@pandacss/core@1.11.3`: `Utility.transform` → `getOrCreateClassName(key, withoutSpace(value))`, no reverse map).

- `class_name_value` no longer scans `values` (drops the dead `literal_matches_class_input` helper).
- `normalize_property_value_cow` no longer substitutes scalar value-map entries onto the atom.
- Authoring the token *key* is unchanged (`css({ marginBottom: '2' })` → `.mb_2`).
- `token()` calls are unaffected — they are evaluated to their CSS string before `css()` sees them, so both the runtime and cssgen name them by the resolved value (`c_oklch(…)`).

```css
/* before */            /* after */
.min-h_screen { … }     .min-h_100vh { min-height: 100vh }
.w_screen { … }         .w_100vw, .w_screen { width: 100vw }   /* both classes now have a rule */
.max-w_screen { … }     .max-w_100vw { max-width: 100vw }
```

## Behavior note

`compiler.atoms()` (a tooling/introspection view) now reports the author's value-map key (`'sm'`, `'full'`, `2`) instead of the pre-resolved CSS. The resolved value is still emitted in declarations; only the introspection surface changed. Two `@pandacss/compiler` snapshots are updated accordingly.

## Tests

- New stylesheet repro `object_map_values_name_classes_by_literal_value` (asserts cssgen names by the authored literal for `marginBottom: '0.5rem'` and `minHeight: '100vh'`).
- Updated the unit/integration snapshots that asserted the old reverse-mapping / pre-resolution.
- `cargo nextest run --workspace` → **1263 passed**; `@pandacss/compiler` Vitest → **151 passed**; `sandbox/codegen` (react) → **89 passed**. `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean on the touched crates.

Refs the class-name divergence discussed in #3599.

### DIFF
--- a/.changeset/fix-classname-value-token-reverse-map.md
+++ b/.changeset/fix-classname-value-token-reverse-map.md
@@ -1,0 +1,11 @@
+---
+'@pandacss/compiler': patch
+---
+
+Fix cssgen naming a class by a token alias the runtime never emits, so the rule silently never matched.
+
+When a utility's `values` is an object map (e.g. preset-base `width`/`minHeight`, whose maps include `screen: '100vw'` / `screen: '100vh'`), the native engine reverse-mapped an authored *resolved value* back to its alias when naming the class: `css({ minHeight: '100vh' })` placed `min-h_100vh` on the element (the runtime hashes `propKey + withoutSpace(value)` verbatim, with no token dictionary) but cssgen wrote only `.min-h_screen`, so `min-height` never applied. Same for `width: '100vw'`, `maxWidth: '100vw'`, and the general object-map case `marginBottom: '0.5rem'` → element `mb_0.5rem`, rule `.mb_2`.
+
+The engine now keeps the author's value-map key on the atom and resolves it to CSS only at emit time, matching the runtime exactly (`min-h_100vh`, `w_100vw`, `mb_0.5rem`) — and matching legacy Panda, which named these by the literal too. Authoring the token *key* is unaffected (`css({ marginBottom: '2' })` → `.mb_2`), and `token()` calls keep their resolved-value class (they are evaluated before `css()`).
+
+Note: `compiler.atoms()` now reports the author's value-map key (e.g. `'sm'`, `'full'`) instead of the pre-resolved CSS — the resolved value is recovered when declarations are written.
--- a/crates/pandacss_project/tests/config.rs
+++ b/crates/pandacss_project/tests/config.rs
@@ -196,12 +196,16 @@ fn utility_values_normalize_aliases() {
     );
 
     assert_eq!(report.css_calls, 1);
+    // Atoms keep the author's value-map *key* (`sm`/`md`), not the resolved CSS.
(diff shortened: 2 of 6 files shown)


segunadebayo:
Thanks for this PR. I've improved it significantly in the v2 branch.

PR #3609, opened 2026-06-19
fix(compiler): keep author order for chained nested `&` selectors (pseudo-relocation)

leegunwoo98:
## Problem

When two nested arbitrary `&` selectors form a parent→descendant chain, v2 relocates a structural pseudo-class off the parent and onto the descendant, producing a selector that matches the wrong elements.

```ts
css({ '&:last-child': { '& .divider': { display: 'none' } } })
```

- **v1 (correct):** `.cls:last-child .divider { display: none }` — `:last-child` qualifies the class-bearing parent; `.divider` is its descendant.
- **v2 (broken):** `.cls .divider:last-child { display: none }` — `:last-child` was relocated onto `.divider`.

Found across ~9 real call sites (e.g. `&:last-child` + `& .requirements-divider`). Same family as the comma-member scope leak (#3607), distinct mechanism.

## Root cause

The cascade sort (`crates/pandacss_stylesheet/src/sort.rs`, `compare_condition_names`) orders selector conditions by pseudo-class priority. For the chain `['&:last-child', '& .divider']` it floats the relational `& .divider` (pseudo-rank 0) ahead of the pseudo `&:last-child` (rank > 0), so the inner descendant is applied first and the pseudo lands on it.

For **named** conditions (`_hover`, `_dark`, …) priority ordering is correct and order-independent. But **raw arbitrary `&` selectors authored as nested object keys** form an authored parent→descendant chain whose order is structurally significant — reordering them changes the ancestor chain.

Verified against legacy `@pandacss/node@1.11.3` (drove its real generator): v1 emits `.cls:last-child .divider`, preserving author order.

## Fix

`compare_condition_names` now returns `Equal` (deferring to the stable sort's author order) when *either* condition is a raw nested-selector key (not a registered breakpoint / container / named / theme condition) that establishes a top-level combinator relationship (descendant ` `, child `>`, adjacent `+`, general sibling `~`). Named conditions keep their existing cascade sort; bare compounds like `&:last-child` still sort normally; the rule grouping/dedup key (`CssRuleKey`, which canonicalizes parts independently) is untouched — only the apply order of nested arbitrary selectors changes.

```css
/* before */ .cls .requirements-divider:last-child { display: none }
/* after  */ .cls:last-child .requirements-divider { display: none }
```

## Tests

- New stylesheet repro `nested_pseudo_then_descendant_keeps_the_pseudo_on_the_parent` asserting `.cls:last-child .divider`.
- Existing `nested_child_selector_stacks_theme_and_direction_conditions` (mixed `& > p` + named theme/direction/hover conditions) is unchanged — confirms named-condition ordering is preserved.
- `cargo nextest run --workspace` → **1263 passed**. `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` clean.

Sibling to #3607 (comma-member scope leak) in the nested-arbitrary-selector family; refs #3599.

### DIFF
--- a/.changeset/fix-nested-selector-pseudo-relocation.md
+++ b/.changeset/fix-nested-selector-pseudo-relocation.md
@@ -1,0 +1,11 @@
+---
+'@pandacss/compiler': patch
+---
+
+Fix a structural pseudo being relocated off the parent onto a nested descendant in chained arbitrary `&` selectors.
+
+Two nested arbitrary-selector object keys form a parent→descendant chain, e.g. `css({ '&:last-child': { '& .divider': { display: 'none' } } })`. v1 (and the intended cascade) emits `.cls:last-child .divider` — the structural pseudo stays on the class-bearing parent. v2 emitted `.cls .divider:last-child`, relocating `:last-child` onto the descendant — a different selector that matches the wrong elements.
+
+The cause is the condition sort: it orders selector conditions by pseudo-class priority, which floats the relational `& .divider` (priority 0) ahead of the pseudo `&:last-child` (priority > 0), inverting the authored nesting. For *named* conditions (`_hover`, `_dark`, …) that ordering is correct and order-independent, but for **raw arbitrary `&` selectors authored as nested object keys** the order is structurally significant.
+
+The sort now preserves author order whenever either side is a raw nested-selector key that establishes a combinator relationship (descendant ` `, child `>`, sibling `+`/`~`). Named/breakpoint/container/theme conditions keep their existing cascade sort, and bare compounds like `&:last-child` still sort normally. The rule grouping/dedup key (`CssRuleKey`) is unaffected — only the apply order of nested arbitrary selectors changes.
--- a/crates/pandacss_stylesheet/src/sort.rs
+++ b/crates/pandacss_stylesheet/src/sort.rs
@@ -401,6 +401,17 @@ impl PartialOrd for SelectorKey {
 }
 
 fn compare_condition_names(config: &UserConfig, a: &str, b: &str) -> Ordering {
+    // Two nested-selector conditions that establish a descendant/child/sibling
+    // relationship are ORDER-SIGNIFICANT: `&:last-child` then `& .divider` must
+    // compose as `.cls:last-child .divider`, never `.cls .divider:last-child`.
+    // The pseudo-rank sort below would float the relational `& .divider`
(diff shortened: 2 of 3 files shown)


leegunwoo98:
### Real-world reproduction on published `2.0.0-beta.1`

Confirmed on a **clean standalone project** with a real `npm i -D @pandacss/dev@2.0.0-beta.1` (not a workspace build) — this PR fixes Shape B (pseudo-relocation).

<details><summary><strong>Repro project</strong> (3 files)</summary>

`panda.config.ts`
```ts
import { defineConfig } from "@pandacss/dev";
export default defineConfig({
  preflight: false,
  include: ["./src/**/*.{ts,tsx}"],
  outdir: "styled-system",
  jsxFramework: "react",
});
```

`src/App.tsx`
```ts
import { css } from "../styled-system/css";

// Shape B — nested structural pseudo on the parent + a descendant target
export const pseudoRelocation = css({
  "&:last-child": {
    "& .child": { display: "none" },
  },
});

// Shape A — comma group with a bare pseudo member (page-blanker, see #3607)
export const commaLeak = css({
  "&:not(:first-child), :only-child": {
    "& .child": { display: "none" },
  },
});
```

Then: `npx panda codegen && npx panda cssgen && cat styled-system/styles.css`
</details>

**`2.0.0-beta.1` output (broken):**
```css
.\[\&\:last-child\]\:\[\&_\.child\]\:display_none .child:last-child, … { display: none }
```

**Shape B (this PR):** for `&:last-child` → `& .child`, the intent is "style `.child` only when the **class-bearing element** is `:last-child`" → v1 emits `.cls:last-child .child`. v2 **relocates** `:last-child` off the class onto the descendant (`.cls .child:last-child`), so it matches whenever `.child` is the last child of *its own* parent, inside *any* class-bearing element — the wrong elements. The class **name** is identical v1↔v2; only the emitted **selector** differs, so a class-name-parity check can't catch it. Real impact in our codebase (~8 sites): collapsed an accordion (first-child `padding-top:0` flush applied to every row), mis-rounded card corners, hid dividers on the wrong rows.

**With this fix**, author order of the chained nested-`&` keys is preserved:
```css
.…:display_none:last-child .child { display: none }
```
(`:last-child` stays on the class; `.child` is its descendant). Matches v1 (`1.11.x`), confirmed by driving the v1 generator directly.

Sibling to #3607 (Shape A / comma-leak) in the nested-arbitrary-selector family.

PR #3634, opened 2026-06-29
feat(designSystem): panda lib + consumer propagation

Adebesin-Cell:
Phase 5 of `designSystem`: the `panda lib` command, plus what a consumer gets from adopting one. Design note: `design-notes/design-system-manifest.md`.

## `panda lib`

Run `panda lib` in a design-system package and it:

- scans your source and writes a portable `panda.buildinfo.json`
- compiles your config into a `preset.mjs` (app and parent fields stripped — `designSystem`, `include`, `outdir`, …)
- writes a `panda.lib.json` manifest that points at both
- syncs your `package.json` exports (`./panda.lib.json`, `./preset`)

It's idempotent — a second run leaves the artifacts byte-identical. `--watch` regenerates on source and config changes.

## Adopting a design system

Set `designSystem: '@acme/ds'` in your config. Panda resolves the manifest, merges its preset, and hydrates its build info into your CSS — so the library's tokens and its pre-extracted styles land in your output without re-scanning its source.

- **Token conflicts.** When a token path is defined by both the library and your config, your value wins and Panda warns. This covers tokens declared under `theme.extend`, not just `theme.tokens`.
- **Stale build info.** When a library's build info can't hydrate (schema or Panda-range mismatch), Panda re-extracts the manifest's `files` instead of failing. It fails closed only when there's no `files` to fall back to.
- **Wrong field.** Put a design system in `include` instead of `designSystem` and Panda flags it and points you to the right field — even though a published library exports only its manifest and preset.

## How it works

- `compilePreset` (`@pandacss/config`) bundles your config through a virtual entry that strips app and parent fields. A nested library ships its own additions; the parent travels in the manifest's `designSystem` field. Functions survive because the module is bundled, not serialized.
- Token-conflict detection (`collectTokenPaths` in `@pandacss/config`) compares your pre-merge token paths against each design-system layer, reading both `theme.tokens`/`semanticTokens` and `theme.extend`.
- The `include` guard and `designSystem` resolution both find a library by probing `${pkg}/panda.lib.json`, so a published package is recognized even though its `exports` expose only the manifest and preset (no `.` or `./package.json`).
- Host-side warnings — stale build info and token conflicts — surface through `Driver.designSystemDiagnostics`.

## Tests

- `@pandacss/config`: `compilePreset`, `lib-manifest`, `token-paths` (base + `extend`), `smart-include` (a `panda lib` package in `include` is flagged).
- `@pandacss/cli`: the `lib` command — artifacts, idempotency, `--watch`.
- `@pandacss/compiler`: consumer hydrate — stale fallback, fail-closed, and token conflict (including `theme.extend`).

## Notes

- Deferred to Phase 6: the styled-system subpath exports (`./css`, `./recipes`, …) that come from codegen. Still open: cross-package source watch and build-info tree-shaking to the consumer's actual imports.


crates/pandacss_stylesheet/tests/.sort.rs.pending-snap
segunadebayo, line 1: this should not be commited

segunadebayo: It seems this PR includes files that should not have been commited. Pls check @Adebesin-Cell

Adebesin-Cell: That was a mistake, reverting 🙏

PR #3652, opened 2026-07-14
feat(config): support minify config key; docs: design system migration path

Adebesin-Cell:
## 📝 Description

Support `minify` as a top-level config key, and document how to publish and consume a design system. From the v2 beta feedback in #3599.

The migration guide and `panda cssgen --minify` (its help reads "overrides config minify") already treated `minify` as a config key, but the `Config` type rejected it and nothing read it — so `minify: true` in `panda.config.ts` failed to typecheck and did nothing. Publishing a design system was also undocumented in the v2 guide.

## 🚀 New behavior

`minify` is a valid top-level config key that `cssgen` reads; the `--minify` flag still overrides it. The v2 migration guide gains a **Design systems** section covering `panda lib`, the artifacts it emits, and the `designSystem` contract — including the trap from the feedback: the build-info file goes in `designSystem`, not `include`. The build-info design note's stale schema version is corrected.

Tests: a `cssgen` case proving config `minify` compacts output without the flag; typecheck and the `cssgen` suite pass.

No.

## 📝 Additional Information

_This PR is #1 — independent of the others._

### DIFF
--- a/.changeset/config-minify-key.md
+++ b/.changeset/config-minify-key.md
@@ -1,0 +1,6 @@
+---
+'@pandacss/types': patch
+'@pandacss/cli': patch
+---
+
+Support `minify` as a top-level config key. The migration guide and `panda cssgen --minify` already treated it as one, but the `Config` type rejected it and nothing read it. `cssgen` now honors `minify` from config, and the `--minify` flag still overrides it.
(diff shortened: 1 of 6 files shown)


Adebesin-Cell:
Folded into #3660 (design-system config & publishing). Closing in favor of the combined PR.

PR #3654, opened 2026-07-15
fix(designSystem): harden build-info hydration (leak, corruption, crash, version skew, option mismatch)

Adebesin-Cell:
## 📝 Description

Harden design-system build-info hydration against the silent-wrong and crashing failure modes the audit surfaced. From the v2 beta feedback in #3599.

Several hydration paths produced wrong or missing CSS, or crashed, without a clear signal:

- `panda lib` in a package that consumes a design system embedded the hydrated parent's atoms in that package's own artifact.
- A corrupt build-info (an out-of-range intern index) silently dropped atoms/recipes while `hydrate` still reported success.
- A structurally invalid but JSON-parseable build-info crashed config load.
- A version- or schema-skewed layer hard-failed the whole build even with a `files` fallback available.
- Overriding `hash`/`prefix`/`separator` away from the design system silently broke its prebuilt class names.

## 🚀 New behavior

- `panda lib` serializes only the package's own extraction; consumers re-hydrate the parent through the chain.
- A corrupt build-info now fails hydration so the consumer re-extracts from `files`.
- A structurally invalid build-info falls back to re-extraction instead of crashing.
- A version/schema-skewed layer re-extracts from `files`.
- You get a warning when your class-name options diverge from the design system's.

Tests: Rust round-trip tests for the parent-atom exclusion and corrupt-table bail; compiler hydrate tests for the version-skew and structural-invalid fallbacks; config tests for the option-mismatch warning. `rust:fmt`/`clippy`, `cargo nextest -p pandacss_project`, the compiler suite, and `typecheck` pass.

No. The `panda lib` artifact shape changes for a package that hydrates a parent, but consumers resolve the parent through the chain, so emitted CSS is unchanged.

## 📝 Additional Information

_This PR is #2 — merge before #3653 (shares `config/design-system.ts`) and #3657 (shares `types/compiler.ts`)._

### DIFF
--- a/.changeset/build-info-hydrate-robustness.md
+++ b/.changeset/build-info-hydrate-robustness.md
@@ -1,0 +1,13 @@
+---
+'@pandacss/compiler': patch
+'@pandacss/compiler-shared': patch
+'@pandacss/config': patch
+---
+
+Harden design-system build-info hydration against silent-wrong and crashing failure modes:
+
+- `panda lib` no longer embeds a hydrated parent's atoms in a middle package's own artifact; a published artifact carries only that package's own extraction.
+- A corrupt build-info (an out-of-range intern index that would silently drop atoms/recipes) is now treated as incompatible, so the consumer re-extracts from `files` instead of hydrating partial CSS.
+- A structurally invalid build-info that still parses as JSON no longer crashes config load; it falls back to re-extraction.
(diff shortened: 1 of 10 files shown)


Adebesin-Cell:
Folded into #3661 (build-info robustness + dead-code removal). Closing in favor of the combined PR.

PR #3655, opened 2026-07-15
fix(compiler): accept multi-major || peer ranges for design systems

Adebesin-Cell:
## 📝 Description

Accept multi-major `||` peer ranges when checking design-system compatibility. From the design-system audit following #3599.

The consumer compatibility check read only the first major from a manifest's `panda` range, so a design system declaring `"panda": "^2.0.0 || ^3.0.0"` rejected consumers on major 3 — a trap that would surface at the v3 cut.

## 🚀 New behavior

The check accepts any major in a `||` union, keeping the fail-closed behavior for unreadable ranges. No new dependency, so it stays wasm-safe.

Tests: multi-major accept/reject cases in `design-system.test.ts`; the existing major-match tests are unchanged.

No — strictly more permissive.

## 📝 Additional Information

_This PR is #4 — merge before #3657 (both touch `compiler-shared/design-system.ts`)._

### DIFF
--- a/.changeset/panda-range-multimajor.md
+++ b/.changeset/panda-range-multimajor.md
@@ -1,0 +1,5 @@
+---
+'@pandacss/compiler-shared': patch
+---
+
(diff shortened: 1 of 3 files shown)


Adebesin-Cell:
Folded into #3660 (design-system config & publishing). Closing in favor of the combined PR.

PR #3663, opened 2026-07-17
feat(cli): add panda studio

Adebesin-Cell:
## 📝 Description

`panda studio` visualizes your design tokens. It replaces the removed v1 Astro studio with two lighter tracks that share one token snapshot:

- `panda studio` — a live viewer built from plain HTML/CSS/JS (no framework, no bundler).
- `panda studio generate` — emits the token views into your project as source, shadcn-style, so you render them wherever you document your design system.

Both read a generated `tokens.json` snapshot. Re-run to refresh after a config change. Design rationale lives in `design-notes/cli-studio-generate.md`.

The v1 studio (`@pandacss/studio`, an Astro app) was removed in the Rust migration. There's no built-in way to look at your resolved tokens.

## 🚀 New behavior

`panda studio` boots a local viewer with a sidebar shell (category nav + scroll-spy + theme toggle), matching the old studio layout but kept light in vanilla HTML/CSS/JS. Each category is presented the way the old studio did:

- colors grouped by family, shades in order
- typography as real sample text at the actual size/weight/font
- spacing and sizes as a scale table (name, value, pixel equivalent, proportional bar)
- radii, borders, shadows, blurs, aspect-ratios as previews
- animations run live and easings animate a dot with their timing function (keyframes are emitted from your config)

It follows your system light/dark theme, has a filter box whose query persists in the URL (`?q=`), and takes `--port`.

`panda studio generate` writes one component per category plus a shared grid into `styled-system/studio` (`--outdir` to change). Views match `config.jsxFramework` (React or Solid), reuse the same presentation, and don't import the styled-system runtime, so they render anywhere. You own the files and re-run to refresh.

No.

## 📝 Additional Information

Notes:

- Generated views are React or Solid. Other frameworks use the vanilla viewer; they can be added later the same way.
- `studio` and `studio generate` are routed as leaf commands in `cli-main` rather than a citty subcommand group, so space-separated flags (`panda studio --port 4000`) parse correctly.
- The contrast checker, typography playground, and semantic tokens all land as stacked PRs on top of this one (see the stack below) — CLI-only, no engine change.

Test plan:

- [x] `pnpm --filter @pandacss/cli test` — studio unit + command tests and help snapshots pass
- [x] `panda studio generate` in `sandbox/storybook` — generated React views typecheck under strict, and a `*.stories.tsx` importing them compiles against the sandbox tsconfig
- [x] `panda studio` against the sandbox — the viewer serves the snapshot and renders the sandbox tokens in light and dark; search filters and round-trips through `?q=`; animations and easings run

## 🧬 Stack

- **this PR** — `panda studio` (base)
- #3664 — contrast checker
- #3665 — typography playground
- #3666 — semantic tokens + named themes

### DIFF
--- a/.changeset/cli-studio.md
+++ b/.changeset/cli-studio.md
@@ -1,0 +1,13 @@
+---
+'@pandacss/cli': minor
+---
+
+Add `panda studio` to visualize your design tokens.
+
+- `panda studio` boots a live viewer — plain HTML/CSS/JS, no framework or bundler — that renders your tokens grouped by
+  category, with light/dark theming. Set the port with `--port`.
+- `panda studio generate` writes the token views into your project as source (shadcn-style) so you own them and render
+  them where you document your design system — an app route, an MDX page, or Storybook. Views match your
+  `config.jsxFramework` (React or Solid) and land in `styled-system/studio` by default (`--outdir` to change).
+
+Both read a generated `tokens.json` snapshot. Re-run to refresh after a config change.
--- a/.prettierignore
+++ b/.prettierignore
@@ -20,3 +20,4 @@ website/pages/**/*.mdx
 .draft/**/*.mdx
 styled-system
 packages/transformer/src/runtime/internal/source.ts
+**/CHANGELOG.md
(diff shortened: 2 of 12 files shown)


Adebesin-Cell:
Consolidated into #3666 — the full `panda studio` feature now ships as one PR.

PR #3666, opened 2026-07-17
feat(cli): add panda studio

Adebesin-Cell:
## 📝 Description

`panda studio` brings back token visualization, gone since the Rust migration removed the v1 Astro Studio. It gives you your resolved tokens in three shapes — data, HTML, and wired-up CSS — plus a live viewer. No framework runtime, no bundler, no shipped styling.

The API follows [Tiptap](https://tiptap.dev)'s `getHTML` / `getJSON`: small functions that hand you one thing each.

v1 shipped Studio as a separate `@pandacss/studio` Astro app: install it, boot a server, self-host to share. The Rust migration removed that pipeline, so there's no built-in way to look at your tokens today.

## 🚀 New behavior

`panda studio` writes `styled-system/studio` into your project and boots a live viewer:

```sh
panda studio                     # emit styled-system/studio + boot the viewer at :4000
panda studio --port 4137
panda studio --css ./studio.css  # render the viewer with your stylesheet
```

`styled-system/studio` exports three framework-agnostic functions:

- **`getTokenJson({ category, query })`** returns your tokens as data. Filter by category, search by query, or omit both for everything. Each token is `{ category, path, name, value, conditions? }`, with values fully resolved and semantic tokens carrying their per-condition and per-theme values.
- **`getTokenHtml({ tokens })`** returns semantic, style-free HTML — names and values as text, everything else on `data-` attributes, all escaped. Drop it into any framework (`dangerouslySetInnerHTML`, `v-html`, `innerHTML`) or write it to a file.
- **`getTokenCss(yourStylesheet)`** takes your CSS and wires it up. It exposes each token's value as a `--pds-value` variable on the matching `[data-value]` element — the one thing static CSS can't derive — then appends your stylesheet. So `var(--pds-value)` resolves to real token values, and values that would break a stylesheet are dropped.

Panda ships the content and the wiring, not the design. There's no default theme and no bundled stylesheet — the grid, swatches, and type are yours. A bare `panda studio` is plain until you pass `--css`; the viewer takes a stylesheet the same way any consumer does.

```tsx
// React
import { getTokenJson, getTokenHtml, getTokenCss } from '../styled-system/studio'
import myCss from './studio.css?raw'

function Colors() {
  const [tokens, setTokens] = useState(getTokenJson({ category: 'colors' }))
  return (
    <div>
      <style>{getTokenCss(myCss)}</style>
      <input onChange={(e) => setTokens(getTokenJson({ category: 'colors', query: e.target.value }))} />
      <div dangerouslySetInnerHTML={{ __html: getTokenHtml({ tokens }) }} />
    </div>
  )
}
```

Full guide: [Panda Studio docs](https://panda-css.com/docs/theming/studio).

No. It's a new, additive command. It replaces the v1 `@pandacss/studio` app, which the Rust migration already removed in v2.

## 📝 Additional Information

The generated module is self-contained: the runtime is inlined with your tokens baked in, so it imports nothing at runtime. Semantic tokens and named themes resolve from `theme.semanticTokens` and `themes`. No Rust changes.


.changeset/cli-studio-contrast.md
segunadebayo, line 1: Let's combine all the 4 changesets into one file

design-notes/cli-studio-v2.md
segunadebayo, line 1: No need for a design note for this one.

packages/cli/src/studio-codegen.ts
segunadebayo, line 1: Looking at this file, it smells like the compiler API should be able to provide some, if not most, of these. So we don't have to maintain this code path

packages/cli/studio-ui/src/showcase.tsx
segunadebayo, line 1: I'm worried this won't always work for all types of user Panda config. It's convenient right now because our config is pretty simple. When things get custom, I'm not sure this will always work.

It's a great idea but let's consider shelving this for now.

website/content/docs/theming/studio.mdx
```diff
@@ -1,42 +1,107 @@
...
+styled-system/studio/
+  tokens.json
+  components/
+    token-grid.tsx
+  Colors.tsx
+  Typography.tsx
+  Spacing.tsx
+  Sizes.tsx
+  Radii.tsx
+  Shadows.tsx
```
segunadebayo: Let's work on the casing of the files to be consistent in the code generation and docs here as well.

Colors.tsx -> colors.tsx

.prettierignore
```diff
@@ -20,3 +20,4 @@ website/pages/**/*.mdx
 .draft/**/*.mdx
 styled-system
 packages/transformer/src/runtime/internal/source.ts
+**/CHANGELOG.md
```
segunadebayo: Any reason for this? Don't think we need it

packages/cli/studio-ui/src/showcase.tsx
Adebesin-Cell: Got it 👍

.prettierignore
Adebesin-Cell: Must have slipped through, reverting...

Adebesin-Cell:
Closing this. Would be superseded by a lighter approach 👍

PR #3673, opened 2026-07-21
fix(engine): run utility transform for shorthands in recipes

Adebesin-Cell:
## 📝 Description

Custom utility `transform` callbacks now run for `cva`/`sva` recipe styles when you author them with the utility's shorthand. Before, the recipe path skipped the transform and emitted the utility key as a literal CSS property, so you got invalid CSS with no build error.

Reported in [#3599](https://github.com/chakra-ui/panda/discussions/3599?sort=new#discussioncomment-17717066).

Take a utility with a `transform` and a `shorthand`:

```ts
utilities: {
  extend: {
    colorVariable: {
      shorthand: 'colorVar',
      className: '--color-var',
      values: 'colors',
      transform: (value) => ({ '--color-var': value }),
    },
  },
}
```

The transform runs in atomic `css()` but not in recipes:

| Usage | Output |
| --- | --- |
| `css({ colorVar: 'red' })` | `--color-var: var(--colors-red)` |
| `cva({ base: { colorVar: 'blue' } })` | `color-variable: var(--colors-blue)` |
| `sva({ slots: ['root'], base: { root: { colorVar: 'green' } } })` | `color-variable: var(--colors-green)` |

`color-variable` is the hyphenated utility key, not the transform's output. It's invalid CSS and no error tells you. This also broke preset-base's own `shadowColor`.

## 🚀 New behavior

Recipes dispatch the transform the same as `css()`, so all three emit `--color-var: var(--colors-…)`.

The cause: atomic `css()` atoms reach the transform already canonicalized (`colorVar` → `colorVariable`), so the callback lookup, which is keyed on the canonical key, hits. Recipe styles keep the shorthand until emit, so the recipe path looked up the raw `colorVar` and missed. The fix resolves the shorthand to the canonical key before dispatch, in both the atomic-atom path (recipe base styles flow through it) and the recipe-entry path (variants and compounds).

No. It fixes output that was already broken. If you depended on the literal `color-variable` declaration, that CSS was never valid.

## 📝 Additional Information

The fix lives in the shared `pandacss_project` crate, so both the native (`@pandacss/compiler`) and wasm (`@pandacss/compiler-wasm`) bindings get it.

Test: `shorthand_in_recipe_dispatches_utility_transform` covers `css`, `cva` base, a `cva` variant, and `sva`. Its transform closure only fires on the canonical key, mirroring the real binding's ref lookup, so a regressed dispatch fails it. `cargo test -p pandacss_stylesheet -p pandacss_project` passes (1021 tests); fmt and clippy are clean.

### DIFF
--- a/.changeset/utility-transform-recipe-shorthand.md
+++ b/.changeset/utility-transform-recipe-shorthand.md
@@ -1,0 +1,6 @@
+---
+'@pandacss/compiler': patch
+'@pandacss/compiler-wasm': patch
+---
+
+Run custom utility `transform` callbacks inside `cva`/`sva` recipes when you use the utility's shorthand. The recipe path dispatched on the raw shorthand, so it skipped the transform and emitted the utility key as a literal CSS property (for example `color-variable: ...` instead of your transform's output). It now resolves the shorthand to the canonical key first, matching atomic `css()`.
--- a/crates/pandacss_project/src/lib.rs
+++ b/crates/pandacss_project/src/lib.rs
@@ -1753,12 +1753,15 @@ fn transform_atoms(
 ) -> FxHashSet<Atom> {
     let mut out = FxHashSet::default();
     for atom in atoms {
-        let resolved = resolved_atom_value(utility, atom.prop(), atom.value());
-        match transform(atom.prop(), &resolved, atom.value()) {
+        let prop = utility.map_or(atom.prop(), |utility| {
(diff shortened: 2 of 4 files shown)


segunadebayo:
The fix is much deeper than just the utility transform fix. I've recently pushed a broader fix across all style creation functions.

PR #3687, opened 2026-07-24
docs: add diagnostics reference

Adebesin-Cell:
## 📝 Description

Adds a Diagnostics reference page: every diagnostic code Panda can emit, what it means, and how to investigate it.

Nothing documents Panda's diagnostics. You see a code like `panda_call_unextractable` or `design_system_export_missing` in the output with no reference for what it means or how to fix it.

## 🚀 New behavior

Adds `references/diagnostics`, listed under References. It covers:

- Errors versus warnings, the exit behavior of each, and `--max-warnings`.
- How diagnostics surface (`--log-level`, `--json`) and how to dig into one with `panda debug`, which links the
  debugging guide.
- Every code grouped by area (config validation, styles at usage, extraction, static CSS, design system, config
  loading), each with its meaning and fix.

Codes and severities are taken from the source: the `pandacss_shared` diagnostic codes plus the config and design-system
diagnostics.

No. Docs only.

### DIFF
--- a/website/content/docs/references/diagnostics.mdx
+++ b/website/content/docs/references/diagnostics.mdx
@@ -1,0 +1,141 @@
+---
+title: Diagnostics
+description: What Panda's diagnostic codes mean, how they surface, and how to investigate them.
+---
+
+Panda reports problems as diagnostics while it loads your config and extracts styles. Each one has a code, a severity,
+and a message. This page explains how they surface and what every code means.
+
+## Errors and warnings
+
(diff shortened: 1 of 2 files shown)


website/content/docs/references/diagnostics.mdx
segunadebayo, line 1: You might need to format this file. The tables look unformatted.

Adebesin-Cell: Resolved

PR #3709, opened 2026-08-06
fix(generator): fix defineParts return type

### LINKED ISSUE #3708: Return type of defineParts is incorrect
### Description

`defineParts` have an incorrect return type. The function returns a callback typed as:

```ts
(config: Partial<Record<keyof T, SystemStyleObject>>) => Partial<Record<keyof T, SystemStyleObject>>
```

This means the returned object is typed with the part keys (e.g. root, label) as keys, and every value is optional regardless of which keys is actually passed in.

That's wrong on two counts:

- __Wrong keys__: at runtime, `defineParts` maps each part key to its `part.selector` value, so the returned object's keys are actually the selectors defined in the `Parts` config, not the original part names.
- __Wrong optionality__: the return type marks every key as optional (Partial<...>), even though the actual returned object only contains the keys that is passed into the config argument.

### Link to Reproduction

See steps to reproduce

### Steps to reproduce

```ts
const parts = defineParts({
  container: { selector: "& [data-container]" },
  item: { selector: "& [data-item]" },
  root: { selector: "&" },
});

const styles = parts({
  item: {
    display: "grid",
    gap: "16",
  },
  root: {
    backgroundColor: "#f00",
  },
});
```

#### Actual return type

```ts
const styles: Partial<Record<"container" | "item" | "root", SystemStyleObject>>
```

#### Expected return type

```ts
const styles: {
 "& [data-item]": SystemStyleObject;
 "&": SystemStyleObject;
}
```

### JS Framework

_No response_

### Panda CSS Version

1.12.0

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

I opened PR #3709 with my proposed fix.

abarani:
Closes #3708

## 📝 Description

Derive the return type from the generic `config` argument (`C`) instead of the static `Parts` type (`T`), and map each key through its `selector`.

The returned type now:

- Only includes keys present in `config` (via `keyof C`)
- Maps each key to its corresponding `selector`, matching the actual runtime output

```ts
const styles: Partial<Record<"container" | "item" | "root", SystemStyleObject>>
```

## 🚀 New behavior

```ts
const styles: {
 "& [data-item]": SystemStyleObject;
 "&": SystemStyleObject;
}
```

I think this is a safe change.

### DIFF
--- a/.changeset/fix-define-parts-return-type.md
+++ b/.changeset/fix-define-parts-return-type.md
@@ -1,0 +1,10 @@
+---
+'@pandacss/generator': patch
+'@pandacss/dev': patch
+---
+
+Fix the return type of `defineParts` so it reflects the actual runtime output.
+
+Calling the returned function now gives you an object typed with the part `selector`s as keys, and only for the parts you
+actually pass in. Previously it was typed as a `Partial` record keyed by the original part names, which was wrong on both
+counts.
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -44,7 +44,9 @@ export function defineSlotRecipe<S extends string, T extends SlotRecipeVariantRe
 }
 
 export function defineParts<T extends Parts>(parts: T) {
-  return function (config: Partial<Record<keyof T, SystemStyleObject>>): Partial<Record<keyof T, SystemStyleObject>> {
+  return <C extends Partial<Record<keyof T, SystemStyleObject>>>(
+    config: C,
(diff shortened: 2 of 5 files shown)


Adebesin-Cell:
Thanks for this, nice catch and a clean fix. I pushed a small follow-up: the `const` type param was only on two of the four `defineParts` declarations. Also dropped in a changeset.

Pulled in @segunadebayo for another set of eyes on the type side.

segunadebayo:
Closing in favor of the note on #3708. Thanks for the PR.

PR #3718, opened 2026-08-13
fix(compiler): replay utility transforms in design-system build info

Adebesin-Cell:
## 📝 Description

A `designSystem` consumer emits the wrong CSS property for any utility with a JS `transform`. A library built with `panda lib` records its extracted styles in `buildinfo.json`, and the consumer replays them instead of re-extracting. That replay dropped the utility's `transform`, so `boxSize="4"` came out as `box-size` instead of `width`/`height` — sometimes a property that doesn't exist.

Reported in [discussion #3599](https://github.com/chakra-ui/panda/discussions/3599#discussioncomment-18000944).

Transform results live in a side map that emit swaps in per usage. `buildinfo.json` never carried that map, and hydrate started it empty, so a replayed atom fell back to the kebab-cased utility name.

```css
/* library's own cssgen — correct */
.size_4 { width: var(--sizes-4); height: var(--sizes-4); }

/* consumer via designSystem — wrong */
.size_4 { box-size: var(--sizes-4); }
```

A transform that returns a condition collapses harder: `.debug_true:hover { border: … }` replays as `.debug_true { debug: true }`. Recipes were unaffected — they bake their transform into the serialized entries at extraction.

## 🚀 New behavior

Each atom carries its transform result in the artifact, and hydrate restores it into the same override map emit reads. A replayed atom produces the CSS the library did, nested conditions and all.

```css
/* consumer via designSystem — now matches the library */
.size_4 { width: var(--sizes-4); height: var(--sizes-4); }
```

No. It's a beta wire-format bump: build info goes to `schemaVersion` 6, and a consumer on this version re-extracts an older artifact from source instead of replaying it. Rebuild your library's build info (`panda lib`) after upgrading.

## 📝 Additional Information

Test plan:

- [x] `cargo nextest run -p pandacss_project build_info` — serialize → hydrate round-trip; the transform styles are gone from the hydrated project without the fix
- [x] `pnpm --filter @pandacss/compiler test` — CSS-level replay of `boxSize` and a nested-condition transform through hydrate (both fail with the restore disabled)
- [x] `pnpm --filter @pandacss/cli test buildinfo`
- [x] `pnpm rust:fmt`, `cargo clippy -p pandacss_project -- -D warnings`

### DIFF
--- a/.changeset/design-system-transform-replay.md
+++ b/.changeset/design-system-transform-replay.md
@@ -1,0 +1,16 @@
+---
+'@pandacss/compiler': patch
+---
+
+Fix a `designSystem` consumer dropping a library utility's `transform` when replaying its build info, emitting the
+wrong (sometimes invalid) property.
+
+```tsx
+// library built with `panda lib`
+<Box boxSize="4" /> // .size_4 { width: …; height: … }
+
+// consumer via `designSystem` — was `.size_4 { box-size: … }`, now width/height
+```
+
+Transform results now travel in the artifact, so a replayed atom emits the same CSS the library did. Requires
+regenerating the library's build info (`schemaVersion` 6); older artifacts re-extract from source.
--- a/crates/pandacss_project/src/build_info.rs
+++ b/crates/pandacss_project/src/build_info.rs
@@ -12,17 +12,17 @@ use pandacss_encoder::{
     Atom, AtomValue, ConditionList, EncodedRecipesSnapshot, RecipeStyleEntry,
     RecipeStyleGroupSnapshot,
 };
-use pandacss_extractor::ExportInfo;
+use pandacss_extractor::{ExportInfo, Literal};
 use pandacss_shared::ViewTransitionStyle;
 use rustc_hash::{FxHashMap, FxHashSet};
 use serde::{Deserialize, Serialize};
 
 use crate::recipes::EncodedRecipes;
-use crate::{FileEntry, ParseFileReport};
+use crate::{FileEntry, ParseFileReport, UtilityStyleKey};
 
 /// Bumped when the on-disk shape changes; a consumer with a different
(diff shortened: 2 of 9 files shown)


segunadebayo:
Diagnosis is right. Closing in favour of `10014b4a8`, which fixes it from the other end.

The consumer already has the library's transform function: `preset.mjs` ships `utilities`, and `createConfigSnapshot` registers `utilities.*.transform` after the preset merge. Your test shows it, `lib` and `app` share one `snapshot()` with callbacks. So we recompute on the consumer instead of shipping the result.

That keeps `schemaVersion` at 5 (no `panda lib` rerun), drops the per-condition duplication (`boxSize: '4'` wrote five identical copies of `s`), and resolves `var(--sizes-4)` against the consumer's tokens.

Both your tests pass unmodified. I added on-disk coverage: real `preset.mjs`, library source stripped, so the rule can only come from hydration.

Still open: a consumer that never merges the preset emits `box-size` silently. Neither approach catches it.

Adebesin-Cell:
> Diagnosis is right. Closing in favour of `10014b4a8`, which fixes it from the other end.
> 
> The consumer already has the library's transform function: `preset.mjs` ships `utilities`, and `createConfigSnapshot` registers `utilities.*.transform` after the preset merge. Your test shows it, `lib` and `app` share one `snapshot()` with callbacks. So we recompute on the consumer instead of shipping the result.
> 
> That keeps `schemaVersion` at 5 (no `panda lib` rerun), drops the per-condition duplication (`boxSize: '4'` wrote five identical copies of `s`), and resolves `var(--sizes-4)` against the consumer's tokens.
> 
> Both your tests pass unmodified. I added on-disk coverage: real `preset.mjs`, library source stripped, so the rule can only come from hydration.
> 
> Still open: a consumer that never merges the preset emits `box-size` silently. Neither approach catches it.

I'll pick up the follow-up task

PR #3739, opened 2026-08-18
ci: comment compiler perf and size on every PR

Adebesin-Cell:
## 📝 Description

Add a benchmark that comments the compiler's speed and output size on every PR, so a change that makes the engine slower or the CSS bigger shows up in review instead of after release.

## 🚀 New behavior

On a pull request that touches the engine (`crates/`, `packages/compiler/`, or `bench/`), a workflow builds the compiler on the base and on the head, runs the same benchmark on both, and posts one sticky comment comparing them.

The benchmark generates a fixed corpus: 200 files, each with a `css` card (four conditions, a responsive query, a nested selector), a `cva` button (two variant axes plus compound variants), and an `sva` menu (a slot recipe). So 200 cards, 200 buttons, and 200 menus each run — enough to move the numbers when the engine changes. The count is set in the workflow if we want more.

The first comment (no base result yet) reads:

| Metric | Base | Head | Change |
| --- | ---: | ---: | ---: |
| `parse.cold.ms` | — | 11.9 | 🆕 new |
| `emit.ms` | — | 2.5 | 🆕 new |
| `css.bytes` | — | 111,090 | 🆕 new |

Once it's on the base branch, later PRs show a real base-vs-head delta with 🟢 / 🔴 / ⚪.

It's meant to be easy to reason about. `parse.cold.ms` reads every file once, `emit.ms` turns that into CSS, `css.bytes` is the output size. It's the median of a few runs, with no network and the same corpus each time, so a move in the numbers is a real move in the engine, not noise.

Both pieces run locally:

- `pnpm --filter=./bench ci-run -- --out result.json --files 200`
- `pnpm --filter=./bench ci-compare -- base.json head.json`

## 📝 Additional Information

- The comment is posted with `actions/github-script` (GitHub's own action) rather than a third-party one, so there's no extra dependency to trust or keep updated.
- The workflow builds the engine twice per run (base and head), since a fair comparison needs both. That's the cost of this first version; it only runs on engine changes, not on every PR.
- The base build tolerates the benchmark script not existing yet, so the first run reports every metric as new.

No.


segunadebayo:
Going to close this for now. The idea is right. This version would cost us two compiler builds on every engine PR, and the timing numbers would not hold up.

The sample in the PR is the problem: `parse.cold.ms` at 11.9 and `emit.ms` at 2.5. A 2% noise floor on 2.5ms is noise on `ubuntu-latest`. Median of 9 runs does not fix that. We'd get 🟢/🔴 we couldn't trust, then ignore the comment.

`css.bytes` is the stable number, and we already guard CSS with stylesheet snapshots. The staticCss / `@container` path already has `static_css_conditions` and `bench/staticcss-build-cost.md`, measured on a named machine.

If we want a PR signal later, size only: bytes, gzip, `@container` count. Leave timing to local hardware.

Adebesin-Cell:
> Going to close this for now. The idea is right. This version would cost us two compiler builds on every engine PR, and the timing numbers would not hold up.
> 
> The sample in the PR is the problem: `parse.cold.ms` at 11.9 and `emit.ms` at 2.5. A 2% noise floor on 2.5ms is noise on `ubuntu-latest`. Median of 9 runs does not fix that. We'd get 🟢/🔴 we couldn't trust, then ignore the comment.
> 
> `css.bytes` is the stable number, and we already guard CSS with stylesheet snapshots. The staticCss / `@container` path already has `static_css_conditions` and `bench/staticcss-build-cost.md`, measured on a named machine.
> 
> If we want a PR signal later, size only: bytes, gzip, `@container` count. Leave timing to local hardware.

Got it, thanks!

PR #3758, opened 2026-08-28
feat(config): suggest the qualified path for a bare token reference

Adebesin-Cell:
## 📝 Description

When a token reference is missing its category, Panda says what's missing but not what to write instead.

```
warning config_token_missing_reference Missing token: `black` used in `theme.semanticTokens.shadows.xl`
```

References resolve by full path from the token root, so a bare `{black}` never resolves — not across categories, not
even inside its own. The value falls through as literal text. `{black/64}` becomes the CSS named color `black`, which
renders, so the page looks fine and the token is quietly ignored.

This comes up when porting a theme written for another system. Chakra's default theme has 8 of these.

## 🚀 New behavior

```
warning config_token_missing_reference Missing token: `black` used in `theme.semanticTokens.shadows.xl`
  help: Did you mean `{colors.black}` or `{fontWeights.black}`? A reference is a full path from the token root, so it
  needs its category.
```

Panda already knows which categories define that name. It lists them, up to two. More than that is too vague to be
worth guessing, so it stays quiet.

No. Same code, same message, same severity — an added `help` line.

## 📝 Additional Information

This is the first Rust-side diagnostic to carry `help`. The field was already plumbed through the NAPI boundary and
the CLI renderer; nothing needed wiring.

Test plan:

- [x] `cargo nextest run -p pandacss_config` (adds three cases: one match, two matches, too many)
- [x] `pnpm rust:test` (2391 tests)
- [x] `pnpm test:compiler`
- [x] ran it against the real config with presets loaded, where `black` matches both `colors` and `fontWeights`

### DIFF
--- a/.changeset/token-ref-suggestion.md
+++ b/.changeset/token-ref-suggestion.md
@@ -1,0 +1,6 @@
+---
+'@pandacss/compiler': patch
+---
+
+Name the fix when a token reference is missing its category. `{black/64}` now warns with
+`Did you mean {colors.black}?` instead of only reporting that `black` is missing.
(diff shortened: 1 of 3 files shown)


segunadebayo:
This is a nice-to-have and can be quite error-prone. It's possible to use `{black/70}` to add opacity to native css color.

Even if we want to consider it, it should be after the release.

PR #3803, opened 2026-09-09
feat(spec-studio): contrast readout, diff view, toolbar polish

Adebesin-Cell:
## 📝 Description

Follow-up UIs for the spec studio, stacked on #3802. This wires up the checker and diff logic that shipped in the base PR, and tweaks the toolbar.

Stacked on `feat/spec-studio-app` — review or merge that first.

The base PR shipped the `contrast` and `diff` utilities with tests, but nothing rendered them. The colors toolbar stacked the heading above the controls.

## 🚀 New behavior

- **Contrast readout** on every color swatch: the token color shown as `Aa` on white and on black, each with its WCAG ratio and level (AA / AAA / Fail). It shows the contrast rather than asking you to pick two colors.
- **Diff view** at `/diff/:slug`: what changed between a spec's last two versions — added, removed, and changed tokens per category, with color swatches. Unchanged categories are left out.
- **Toolbar**: the filter input is larger and the actions sit on the top row, spaced across from the heading.

No.

## 📝 Additional Information

- The diff view compares the latest version against the one before it — no version picker yet. A link into it from the shared view is a small follow-up.
- Still to come from the parity list: a named-`themes` selector, Text/Layer Styles, and a typography composer.

### Test plan

- [x] `pnpm --filter @pandacss/spec-studio test` — passes
- [x] Contrast readout: rendered a 5-color system, checked the ratios against white/black (amber fails on white, passes on black; slate the reverse)
- [x] Diff view: published a v2 with an added, a removed, and a changed color plus a new category — all render correctly, unchanged category omitted


Adebesin-Cell:
Folded into #3802 — everything ships as one PR now.

